PR #30, opened 2017-04-03
v1.2.0 - adding support of different RF modules

WoodsterDK:
Added new RF interface, but still some minor issues:

- web app does not store interface type
- web app crashes on submit, might very well be the addition of the radio interface type
- the actual transmit on the PL1167 interface sometimes does not come through,
will look at it

### DIFF
--- a/data/web/index.html
+++ b/data/web/index.html
@@ -16,7 +16,7 @@
   <!--[if lt IE 9]>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/html5shiv/3.7.3/html5shiv.js"></script>
   <![endif]-->
-  
+
   <style>
     .header-row { border-bottom: 1px solid #ccc; }
     label { display: block; }
@@ -31,8 +31,8 @@
     .error-info:before { content: '('; }
     .error-info:after { content: ')'; }
     .header-btn { margin: 20px; }
-    .btn-secondary { 
-      background-color: #fff; 
+    .btn-secondary {
+      background-color: #fff;
       border: 1px solid #ccc;
     }
     .inline { display: inline-block; }
@@ -46,41 +46,41 @@
       width: calc(100% - 3em);
       display: inline-block;
       cursor: pointer;
-      background: linear-gradient(to right, 
-        rgb(255, 0, 0) 0%, 
-        rgb(255, 255, 0) 16.6667%, 
-        rgb(0, 255, 0) 33.3333%, 
-        rgb(0, 255, 255) 50%, 
-        rgb(0, 0, 255) 66.6667%, 
-        rgb(255, 0, 255) 83.3333%, 
+      background: linear-gradient(to right,
+        rgb(255, 0, 0) 0%,
+        rgb(255, 255, 0) 16.6667%,
+        rgb(0, 255, 0) 33.3333%,
+        rgb(0, 255, 255) 50%,
+        rgb(0, 0, 255) 66.6667%,
+        rgb(255, 0, 255) 83.3333%,
         rgb(255, 0, 0) 100%
       );
     }
-    .hue-value-display { 
+    .hue-value-display {
       border: 1px solid #000;
       margin-left: 0.5em;
       width: 2em;
       height: 2em;
       display: inline-block;
     }
-    .plus-minus-group { 
+    .plus-minus-group {
       overflow: auto;
       width: 100%;
       clear: both;
       display: block;
     }
-    .plus-minus-group button:first-of-type { 
+    .plus-minus-group button:first-of-type {
       border-bottom-right-radius: 0;
       border-top-right-radius: 0;
       float: left;
       display: block;
     }
-    .plus-minus-group button:last-of-type { 
+    .plus-minus-group button:last-of-type {
       border-bottom-left-radius: 0;
       border-top-left-radius: 0;
       display: block;
     }
-    .plus-minus-group .title { 
+    .plus-minus-group .title {
       border-width: 1px 0;
       border-color: #ccc;
       border-style: solid;
@@ -100,12 +100,12 @@
       animation: spin 1s infinite linear;
       -webkit-animation: spin2 1s infinite linear;
     }
-    
+
     @keyframes spin {
       from { transform: scale(1) rotate(0deg); }
       to { transform: scale(1) rotate(360deg); }
     }
-    
+
     @-webkit-keyframes spin2 {
       from { -webkit-transform: rotate(0deg); }
       to { -webkit-transform: rotate(360deg); }
@@ -120,14 +120,15 @@
   <script src="https://cdnjs.cloudflare.com/ajax/libs/underscore.js/1.8.3/underscore-min.js"></script>
   <script src="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-slider/9.7.0/bootstrap-slider.min.js"></script>
   <script src="https://cdnjs.cloudflare.com/ajax/libs/selectize.js/0.12.4/js/standalone/selectize.min.js"></script>
-  
+
   <script lang="text/javascript">
     var FORM_SETTINGS = [
-      "admin_username", "admin_password", "ce_pin", "csn_pin", "packet_repeats",
+      "admin_username", "admin_password", "ce_pin", "csn_pin", "reset_pin","packet_repeats",
       "http_repeat_factor", "auto_restart_period"
     ];
-    
+
     var FORM_SETTINGS_HELP = {
+      csn_pin_RxReady : "'csn' for 'nRF24' interface, and RxReady for 'PL1167_LT8900' interface",
       packet_repeats : "The number of times to repeat RF packets sent to bulbs",
       http_repeat_factor : "Multiplicative factor on packet_repeats for " +
         "requests initiated by the HTTP API. UDP API typically receives " +
@@ -135,37 +136,37 @@
       auto_restart_period : "Automatically restart the device every number of " +
         "minutes specified. Use 0 for disabled."
     }
-    
+
     var UDP_PROTOCOL_VERSIONS = [ 5, 6 ];
     var DEFAULT_UDP_PROTOCL_VERSION = 5;
-    
+
     var selectize;
-    
+
     var toHex = function(v) {
       return "0x" + (v).toString(16).toUpperCase();
     }
-  
+
     var activeUrl = function() {
       var deviceId = $('#deviceId option:selected').val()
         , groupId = $('#groupId .active input').data('value')
         , mode = getCurrentMode();
-        
+
       if (deviceId == "") {
         alert("Please enter a device ID.");
         throw "Must enter device ID";
       }
-      
+
       if (! $('#group-option').data('for').split(',').includes(mode)) {
         groupId = 0;
       }
-        
+
       return "/gateways/" + deviceId + "/" + mode + "/" + groupId;
     }
-    
+
     var getCurrentMode = function() {
       return $('input[name="mode"]:checked').data('value');
     };
-    
+
     var updateGroup = _.throttle(
       function(params) {
         $.ajax(
@@ -179,7 +180,7 @@
       },
       1000
     );
-    
+
     var sendCommand = _.throttle(
       function(params) {
         $.ajax(
@@ -193,18 +194,18 @@
       },
       1000
     )
-    
+
     var sniffRequest;
     var sniffing = false;
     var getTraffic = function() {
       var sniffType = $('#sniff-type input:checked').data('value');
-      
+
       sniffRequest = $.get('/gateway_traffic/' + sniffType, function(data) {
         $('#sniffed-traffic').html(data + $('#sniffed-traffic').html());
         getTraffic();
       });
     };
-    
+
     var gatewayServerRow = function(deviceId, port, version) {
       var elmt = '<tr>';
       elmt += '<td>';
@@ -215,17 +216,17 @@
       elmt += '</td>';
       elmt += '<td>';
       elmt += '<div class="btn-group" data-toggle="buttons">';
-      
+
       for (var i = 0; i < UDP_PROTOCOL_VERSIONS.length; i++) {
         var val = UDP_PROTOCOL_VERSIONS[i]
           , selected = (version == val || (val == DEFAULT_UDP_PROTOCL_VERSION && !UDP_PROTOCOL_VERSIONS.includes(version)));
-        
+
         elmt += '<label class="btn btn-secondary' + (selected ? ' active' : '') + '">';
-        elmt += '<input type="radio" name="versions[]" autocomplete="off" data-value="' + val + '" ' 
+        elmt += '<input type="radio" name="versions[]" autocomplete="off" data-value="' + val + '" '
           + (selected ? 'checked' : '') +'> ' + val;
         elmt += '</label>';
       }
-      
+
       elmt += '</div></td>';
       elmt += '<td>';
       elmt += '<button class="btn btn-danger remove-gateway-server">';
@@ -235,17 +236,17 @@
       elmt += '</tr>';
       return elmt;
     }
-    
+
     var loadSettings = function() {
       $.getJSON('/settings', function(val) {
         Object.keys(val).forEach(function(k) {
           var field = $('#settings input[name="' + k + '"]');
-          
+
           if (field.length > 0) {
             field.val(val[k]);
           }
         });
-        
+
         if (val.device_ids) {
           selectize.clearOptions();
           val.device_ids.forEach(function(v) {
@@ -253,7 +254,7 @@
           });
           selectize.refreshOptions();
         }
-        
+
         var gatewayForm = $('#gateway-server-configs').html('');
         if (val.gateway_configs) {
           val.gateway_configs.forEach(function(v) {
@@ -262,16 +263,16 @@
         }
       });
     };
-    
+
     var saveGatewayConfigs = function() {
       var form = $('#gateway-server-form')
         , errors = false;
-        
+
       $('input', form).removeClass('error');
-      
+
       var deviceIds = $('input[name="deviceIds[]"]', form).map(function(i, v) {
         var val = $(v).val();
-        
+
         if (isNaN(val)) {
           errors = true;
           $(v).addClass('error');
@@ -280,10 +281,10 @@
           return val;
         }
       });
-      
+
       var ports = $('input[name="ports[]"]', form).map(function(i, v) {
         var val = $(v).val();
-        
+
         if (isNaN(val)) {
           errors = true;
           $(v).addClass('error');
@@ -292,11 +293,11 @@
           return val;
         }
       });
-      
+
       var versions = $('.active input[name="versions[]"]', form).map(function(i, v) {
         return $(v).data('value');
       });
-        
+
       if (!errors) {
         var data = [];
         for (var i = 0; i < deviceIds.length; i++) {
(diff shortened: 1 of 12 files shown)


sidoh:
Awesome. Thanks for putting this together!

Are you at the point where this is mostly finalized? I have some questions/suggestions, but don't want to jump the gun if you're still iterating.

WoodsterDK:
It works, only missing part is the save of selected interface in the app.
Please feel free to ask or come with suggestions.

sidoh:
Sweet. :)

WoodsterDK:
I havent been able to test that the nRF interface still works.
So far you need to manually change in the settings file. Can you verify that it still works?

sidoh:
Yeah, I'll definitely test with the nRF. Want to test with the one you're adding as well, but those are probably still a few weeks out. Do you know of a place to get them that ships faster than Chinese ebay suppliers?

I'll take a crack at reviewing later today.

WoodsterDK:
Super, no not really, ebay or aliexpress are what I have used.
There is still an issue saving the selected interface. Perhaps you can give the javascript a look, and complete it.
You can try to measure the time used for transmitting one command (repeats included) then we can see if there is any speed to gain switching to a LT8900 interface.

WoodsterDK:
Perhaps here, dont know if they ship, and quite pricey compared to ebay.
https://authometion.com/shop/en/wifi-and-radio-modules/6-pl1167-radio-module.html

lib/MiLight/MiLightRadioPL1167_LT8900.h
```diff
@@ -0,0 +1,95 @@
+/*
+ * MiLightRadio.h
+ *
+ *  Created on: 29 May 2015
+ *      Author: henryk
+ */
```
sidoh: Can we delete this header?

lib/MiLight/MiLightRadioInterface.h
```diff
@@ -0,0 +1,29 @@
...
+#include "Arduino.h"
+#else
+#include <stdint.h>
+#include <stdlib.h>
+#endif
+
+#ifndef MILIGHTRADIOINTERFACE_H_
+#define MILIGHTRADIOINTERFACE_H_
+
+class MiLightRadioInterface
```
sidoh: What do you think about the name `RadioProvider`? I'm finding it a little confusing that there's a `MiLightRadioInterface` with an impl named `MiLightRadio` (maybe just a C++ pattern that I'm not familiar with?)

lib/MiLight/MiLightRadio.h
```diff
@@ -15,24 +15,25 @@
...
 #include <MiLightRadioConfig.h>
+#include <MiLightRadioInterface.h>
 
 // #define DEBUG_PRINTF
 
 #ifndef MILIGHTRADIO_H_
 #define MILIGHTRADIO_H_
 
-class MiLightRadio {
+class MiLightRadio : public MiLightRadioInterface{
```
sidoh: Should we rename this to contain something to indicate it's for the NRF?

lib/MiLight/RadioStack.h
```diff
@@ -2,33 +2,55 @@
...
 #include <MiLightRadio.h>
+#include <MiLightRadioPL1167_LT8900.h>
+#include <MiLightRadioInterface.h>
+#include <Settings.h>
 
 #ifndef _RADIO_STACK_H
-#define _RADIO_STACK_H 
+#define _RADIO_STACK_H
 
 class RadioStack {
```
sidoh: I'm thinking that `RadioStack` was and is kind of serving the role of a poorly named and half-heartedly implemented Factory. Its original purpose was to contain the logic to construct an instance of `MiLightRadio`, but it's not doing a very good job of that since you still have to pass in an instance of `RF24`.

I think if we renamed this `RadioProviderFactory` and had two impls that took in only the radio config and the hardware config (pins, etc.), we could just pass that into `MiLightClient` and the separation would be a whole lot cleaner.

lib/Settings/Settings.cpp
```diff
@@ -31,34 +31,51 @@ void Settings::deserialize(Settings& settings, JsonObject& parsedSettings) {
...
+    }
+
+    if (parsedSettings.containsKey("radio_interface_type"))
+    {
+      String InterfaceType = parsedSettings["radio_interface_type"];
+      if(InterfaceType == "nRF24")
+      {
+        settings.radioInterfaceType =  nRF24;
+      }
+      else if(InterfaceType == "PL1167/LT8900/8910/8920")
```
sidoh: Does this thing have a standard name? Is one of these the "real" name and others are resold under different model numbers? I think it's fine to use this name to display in the UI, but I'm thinking it should probably have a tighter representation in code.

Might be nice to pull this into a helper since it's used more than once.

lib/Settings/Settings.h
```diff
@@ -45,6 +51,8 @@ class Settings {
     // CE and CSN pins from nrf24l01
     cePin(D0),
     csnPin(D8),
+    resetPin(0),
+    radioInterfaceType(PL1167_LT8900),
```
sidoh: Think we should default to NRF, for the sake of backwards compatibility if nothing else.

platformio.ini
```diff
@@ -10,23 +10,25 @@
...
-lib_deps_builtin = 
+lib_deps_builtin =
   SPI
-lib_deps_external = 
+lib_deps_external =
   RF24
   WiFiManager
   ArduinoJson
-build_flags = !echo -D MILIGHT_HUB_VERSION=\\\"$(git describe --always)\\\"
+build_flags = !echo -D MILIGHT_HUB_VERSION="\"$(git describe --always)\""
```
sidoh: Was the extra string escaping breaking for you?  What platform are you on?  It seemed necessary to double-escape on both OS X and Linux (Travis).

platformio.ini
```diff
@@ -10,23 +10,25 @@
...
+; build_flags = -D MILIGHT_HUB_VERSION=\"1.1.0\" -D MILIGHT_UDP_DEBUG
 # -D MILIGHT_UDP_DEBUG
+upload_speed = 9600
```
sidoh: Was flashing not working with the default baudrate? What board are you using? Looks like d1_mini? I've flashed a few different d1_minis with the default baudrate without any trouble.

platformio.ini
```diff
@@ -10,23 +10,25 @@
...
 
-[env:nodemcuv2]
-platform = espressif8266
-framework = arduino
-board = nodemcuv2
-lib_deps =
-  ${common.lib_deps_builtin}
-  ${common.lib_deps_external}
-build_flags = ${common.build_flags} -D FIRMWARE_VARIANT=\\\"nodemcuv2\\\"
+; [env:nodemcuv2]
```
sidoh: Can we uncomment this? If you want to build for a specific environment, you can do `platformio run -e d1_mini`.

lib/MiLight/MiLightRadioPL1197_LT8900.cpp
```diff
@@ -0,0 +1,631 @@
...
+		digitalWrite(byResetPin, LOW);
+		delay(200);
+		digitalWrite(byResetPin, HIGH);
+		delay(200);
+	}
+
+  pinMode(_csPin, OUTPUT);
+  digitalWrite(_csPin, HIGH);
+
+  SPI.begin();
```
sidoh: Should this setup stuff be in `begin()`?

lib/MiLight/MiLightRadioPL1197_LT8900.cpp
```diff
@@ -0,0 +1,631 @@
...
+/**************************************************************************/
+// Initialize radio module
+/**************************************************************************/
+void MiLightRadioPL1167_LT8900::vInitRadioModule(MiLightRadioType type)
+{
+	if (type == RGB_CCT)
+	{
+		bool bWriteDefaultDefault = true;  // Is it okay to use the default power up values, without setting them
+
+		regWrite16(0x00, 0x6F, 0xE0, 7);  // Recommended value by PMmicro
```
sidoh: Looks like a lot of these values are the same for the other radio types. Do you know how much of it can be collapsed?

lib/MiLight/MiLightRadioPL1197_LT8900.cpp
```diff
@@ -0,0 +1,631 @@
...
+          Serial.println(F(" Frame: "));
+          //dump the packet.
+
+          _packet[0]=byaFramesSizes[byCounterFrame];
+          for (int iByteCounter = 0; iByteCounter < byaFramesSizes[byCounterFrame]; iByteCounter++)
+          {
+            byte byReceivedvalue = byaFramesReceived[byCounterFrame][iByteCounter];
+            _packet[iByteCounter+1] = byReceivedvalue;
+
+            String stringOne = String(byReceivedvalue, HEX);
```
sidoh: Might recommend using `printf("%02X ", byReceivedvalue)` here instead.

lib/MiLight/MiLightRadioPL1197_LT8900.cpp
```diff
@@ -0,0 +1,631 @@
...
+      if (byFrameCounter != 0)
+      {
+        Serial.println("");
+        Serial.print(F("Packets received: "));
+        Serial.print(byFrameCounter);
+
+        for (byte byCounterFrame = 0; byCounterFrame < byFrameCounter; byCounterFrame++)
+        {
+          _dupes_received = byaFramesReceivedCount[byCounterFrame];
+          Serial.println("");
```
sidoh: Can we wrap this and other prints here in `#ifdef DEBUG_PRINTF` or some other toggleable debug flag?

lib/MiLight/MiLightRadioPL1197_LT8900.cpp
```diff
@@ -0,0 +1,631 @@
...
+ *
+ *  Created on: 31 March 2017
+ *      Author: WoodsterDK
+ *
+ *  Very inspired by:
+ *  https://github.com/pmoscetta/authometion-milight/tree/master/Authometion-MiLight
+ *  https://bitbucket.org/robvanderveer/lt8900lib
+ */
+
+#include "MiLightRadioPL1167_LT8900.h"
```
sidoh: The formatting here is a little inconsistent. Can we run some autoformatter with:

* K&R braces (I know this is a personal preference, but would be nice for it to be consistent within the project)
* Two space indents

Obviously I'm happy to do this within a branch if you'd prefer not to.

lib/MiLight/MiLightClient.cpp
```diff
@@ -2,40 +2,42 @@
...
       radio->configure();
     }
-    
+
     currentRadio = stack;
     formatter = stack->config.packetFormatter;
     return radio;
   } else {
-    Serial.print(F("MiLightClient - tried to get radio for unknown type: "));
+    Serial.print("MiLightClient - tried to get radio for unknown type: ");
```
sidoh: Was the `F` PROGMEM macro not working here?

data/web/index.html
```diff
@@ -521,53 +522,70 @@
...
-        
-        elmt += '<input type="text" class="form-control" name="' + k + '"/>';
-        elmt += '</div>';
-        
+
+        if(k === "radio_interface_type")
+        {
+          elmt += '<div class="btn-group" id="radiointerface" data-toggle="buttons">' +
+            '<label class="btn btn-secondary active">' +
+            '<input type="radio" name="radiointerface_nRF" autocomplete="off" data-value="nRF24" > nRF24' +
```
sidoh: The two inputs should have the same `name`, right?

data/web/index.html
```diff
@@ -576,17 +594,19 @@
             data: JSON.stringify(obj)
           }
         );
-        
+        //console.log(obj);
```
sidoh: can we remove this?

lib/MiLight/MiLightClient.cpp
```diff
@@ -45,41 +47,45 @@ void MiLightClient::setResendCount(const unsigned int resendCount) {
...
   if (currentRadio == NULL) {
     return;
   }
-  
-  size_t length;
-  currentRadio->getRadio()->read(packet, length);
+
+  size_t length = currentRadio->config.getPacketLength();
+
+  currentRadio->getRadioInterface()->read(packet, length);
```
sidoh: I think the interface here is supposed to be that `length` is passed as a reference, and is altered to be the length of the read packet.

We can probably just remove it. It's an artifact from Henryk's code, and I don't think it's useful anywhere.

lib/MiLight/MiLightClient.cpp
```diff
@@ -45,41 +47,45 @@ void MiLightClient::setResendCount(const unsigned int resendCount) {
...
+
 #ifdef DEBUG_PRINTF
   printf("Sending packet: ");
   for (int i = 0; i < currentRadio->config.getPacketLength(); i++) {
     printf("%02X", packet[i]);
   }
   printf("\n");
 #endif
-  
+  int iStart = millis();
```
sidoh: Can we wrap this in a `#ifdef` debug flag?

sidoh:
Also obviously the travis build should be passing :)

I think it's just failing now because of the changes to `platformio.ini`.

WoodsterDK:
You were right regarding the travis build - the \\\"$(git describe --always)\\\"  and FIRMWARE_VARIANT=\\\"nodemcuv2\\\" gives me an compile error on a windows environment, that is why I changed...

lib/MiLight/MiLightClient.h
```diff
@@ -14,38 +16,53 @@
 
 class MiLightClient {
   public:
-    MiLightClient(uint8_t cePin, uint8_t csnPin)
-      : rf(RF24(cePin, csnPin)),
-      resendCount(MILIGHT_DEFAULT_RESEND_COUNT),
-      currentRadio(NULL),
-      numRadios(MiLightRadioConfig::NUM_CONFIGS)
+  MiLightClient(byte byCSPin, byte byResetPin, byte byPktFlag, eRadioInterfaceType InterfaceType)
```
WoodsterDK: Good idea, - I went after prove of concept - feel free mock around with the code.

lib/MiLight/MiLightClient.h
```diff
@@ -14,38 +16,53 @@
...
+  {
+    radios = new RadioStack*[numRadios];
+
+    for (size_t i = 0; i < numRadios; i++)
     {
-      radios = new RadioStack*[numRadios];
-      
-      for (size_t i = 0; i < numRadios; i++) {
-        radios[i] = new RadioStack(rf, *MiLightRadioConfig::ALL_CONFIGS[i]);
+      if(InterfaceType == PL1167_LT8900)
```
WoodsterDK: Good idea, - I went after prove of concept - feel free mock around with the code.

lib/MiLight/MiLightRadioPL1167_LT8900.h
WoodsterDK: Yes.... : )

lib/MiLight/MiLightRadioInterface.h
WoodsterDK: Good idea, - I went after prove of concept - feel free mock around with the code.

lib/MiLight/MiLightRadio.h
WoodsterDK: Good idea, - I just didn't want to mess too much around with your code.

lib/MiLight/RadioStack.h
WoodsterDK: Good idea - the RF24 instance actually irritated me there, but again, I didn't wanted to mess too much around with your code.

lib/Settings/Settings.cpp
WoodsterDK: True - Lstt2005 came up with the description for the UI, and it just stuck. Don't know if there is a standard. Might just be "PL1167_Compatible"?

lib/Settings/Settings.h
WoodsterDK: Sure thing - I left it according to my settings. No problem.

lib/MiLight/MiLightRadioPL1197_LT8900.cpp
WoodsterDK: Sure thing.

lib/MiLight/MiLightRadioPL1197_LT8900.cpp
WoodsterDK: No - originally the settings came from https://github.com/pmoscetta/authometion-milight.
For the RGBWW_CCT I read the settings using a logic analyzer - don't know why they are different.

lib/MiLight/MiLightRadioPL1197_LT8900.cpp
WoodsterDK: Normally I initialize in the constructor.

lib/MiLight/MiLightRadioPL1197_LT8900.cpp
WoodsterDK: Yes.

lib/MiLight/MiLightClient.cpp
WoodsterDK: It should - I think I started the code before you made the PROGMEM macro addition. That is why it's different.

lib/MiLight/MiLightClient.cpp
WoodsterDK: Sure thing

data/web/index.html
WoodsterDK: Yes - that might be why it is not working. If you can fix the write then super.

data/web/index.html
WoodsterDK: Sure thing - I tried to get it to work.

lib/MiLight/MiLightRadioPL1197_LT8900.cpp
WoodsterDK: I absolutely know what you mean, normally I can't stand different layouts - but here I went for a prof of concept, and again didn't want to mess around too much in you code.
If you like please change according to your preferences.
I'm happy to help with your super project - and feel the code inserted should be something you can accept and up to your standard. (Absolutely no hard feelings : ))

WoodsterDK:
Did you try to do a speed test on the nRF interface?

sidoh:
Cool - I'm happy to run in this direction if you're comfortable with it. I wanted to be really specific with what I was imagining in case:

1. You knew of a reason it wouldn't work
1. You had different preferences
1. You wanted to merge stuff in without me fiddling with it

Sorry if I'm coming across as pedantic, just wanted to make sure you were cool with the changes :)

I'll run a speed test in a sec. I was soldering a prototype board so I could ditch the dupont connectors.

sidoh:
Also dammit @ the build not working on Windows. I was afraid of that. I'll have to try making it more platform independent. :(

WoodsterDK:
Speed test - it should actually be printed out in serial as the code is now - the time is calculated void "MiLightClient::write(uint8_t packet[])"

Not pedantic at all - again it is your very impressive project, and I'm happy to chip in.
I prefer you to arrange it as you like.

sidoh:
Cool, I'll give that a shot!

Sounds great. I'll give this a go and merge it into a branch.

Thanks again for digging into this. I'm excited to try the different module.

lib/MiLight/MiLightRadioPL1197_LT8900.cpp
sidoh: Interesting. I was asking because the 4 or 5 I spot-checked looked like they were the same (e.g., it looks like 0x00 -- 0x07 are the same). Do you know offhand which values are different?

PR #38, opened 2017-04-05
Speed optimisation and bug fixes

WoodsterDK:
Speed optimisation and bug fixes:
- Changes to dual module has been verified to work.

### DIFF
--- a/lib/MiLight/MiLightRadioFactory.cpp
+++ b/lib/MiLight/MiLightRadioFactory.cpp
@@ -6,7 +6,7 @@ MiLightRadioFactory* MiLightRadioFactory::fromSettings(const Settings& settings)
       return new NRF24Factory(settings.csnPin, settings.cePin);
 
     case LT8900:
-      return new LT8900Factory(settings.csnPin, settings.resetPin, settings.cePin);
+      return new LT8900Factory(settings.cePin, settings.resetPin, settings.csnPin);
 
     default:
       return NULL;
--- a/lib/MiLight/LT8900MiLightRadio.cpp
+++ b/lib/MiLight/LT8900MiLightRadio.cpp
@@ -38,10 +38,10 @@ LT8900MiLightRadio::LT8900MiLightRadio(byte byCSPin, byte byResetPin, byte byPkt
 
   SPI.begin();
 
-  SPI.setFrequency(1000000);
   SPI.setDataMode(SPI_MODE1);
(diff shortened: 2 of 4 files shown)


lib/MiLight/MiLightRadioFactory.cpp
```diff
@@ -6,7 +6,7 @@ MiLightRadioFactory* MiLightRadioFactory::fromSettings(const Settings& settings)
       return new NRF24Factory(settings.csnPin, settings.cePin);
 
     case LT8900:
-      return new LT8900Factory(settings.csnPin, settings.resetPin, settings.cePin);
+      return new LT8900Factory(settings.cePin, settings.resetPin, settings.csnPin);
```
sidoh: I apologize in advance for being dense here. Just wanna make sure I understand:

It's a little funky that this is the correct order when the arguments are passed in through this path:

```c++
  LT8900Factory(uint8_t csPin, uint8_t resetPin, uint8_t pktFlag);

...

MiLightRadio* LT8900Factory::create(const MiLightRadioConfig& config) {
  return new LT8900MiLightRadio(_csPin, _resetPin, _pktFlag, config);
}

...

    LT8900MiLightRadio(byte byCSPin, byte byResetPin, byte byPktFlag, const MiLightRadioConfig& config);
```

Shouldn't the way it is be the correct order? It's a little confusing that what's in the "ce" setting eventually gets passed into `byCSPin`. If you flipped the setting values for your cs/ce pins without this change, would you get the same result?

sidoh:
Awesome! :D

WoodsterDK:
Regarding the pins, the module I have only has a combined select/enable.
If you like, you can switch them, i will just revert in the web config. It is ok.
I just interpreted it as shown.

sidoh:
Aah, gotcha. I see what you're saying.

It seems like the pin is labeled "CS" on the device. Is that right? If that's the case, maybe it makes sense to bind the "csn" setting to this pin and have "CE" be the floating setting that's CE on the NRF and pktFlag on the LT8900. Maybe there should even be different pin settings shown depending on what board is selected.

WoodsterDK:
Sounds OK with your marking.
Isn't different pin settings too much work compared to the benefit. Maybe describe it in the help text?

sidoh:
👍 

Yeah, you're probably right. I'm certainly leaning that way too. I guess I just don't want it to be super confusing to hook up the radio :)

sidoh:
I'll merge and flip the settings around and update the UI / docs later.

PR #69, opened 2017-04-25
Fixed the switched ON/OFF for {all}

phigmeta:
If you send the UDP signal for "ALL ON" (42 00 55 - All On) it turns everything off

If you Send the UDP signal for "ALL OFF" (41 00 55 - All Off) it turns everything on

Changed the names around to keep the numbering in order.


sidoh:
Thank you!

PR #80, opened 2017-05-10
added a system command to reset the wifi and then restart the device

nothingmn:


sidoh:
Nice! :)

If you don't mind, I'm going to point this at the v1.3.0 branch, which I'll merge into master in the next week or so.

lib/WebServer/MiLightHttpServer.cpp
```diff
@@ -128,7 +128,19 @@ void MiLightHttpServer::handleSystemPost() {
       delay(100);
 
       ESP.restart();
-    }
+
+      handled = true;
+    } else if (request["command"] == "reset_wifi") {
```
sidoh: I wonder if a name like `reset_wifi_config` might be clearer? `reset_wifi` makes it sound like it might just be re-connecting to the AP.

nothingmn:
All good.  I was going to add a button to the UI to light it up as well.

sidoh:
Nice, thanks for adding the button. What do you think about this comment?

> I wonder if a name like `reset_wifi_config` might be clearer? `reset_wifi` makes it sound like it might just be disconnecting and re-connecting to the AP.

nothingmn:
Being more explicit with the name is definitely a good idea.  Id also support  

clear_wifi_config
kill_wifi_settings
obliterate_wifi_bits

:)

sidoh:
> obliterate_wifi_bits

hahaha :)

`clear_wifi_config` sounds really clear to me. You able to make that change to the command and the button?

nothingmn:
Updated, also updated the text on the button to be just as clear.  Finally I added the meta tag to the head, for better rendering on mobile phones.

Cheers.

lib/Settings/Settings.h
```diff
@@ -21,7 +21,7 @@
 
 #define WEB_INDEX_FILENAME "/web/index.html"
 
-#define MILIGHT_GITHUB_USER "sidoh"
+#define MILIGHT_GITHUB_USER "nothingmn"
```
sidoh: Guessing this snuck in without you meaning for it to. Can we change it back? :)

sidoh:
Awesome, looks great. Thanks! 

Can merge to the 1.3 branch after we revert the `Settings.h` change :)

lib/Settings/Settings.h
nothingmn: ack.  good catch.  already pushed up the change.

btw.  whats the easiest way to test out UI changes?

sidoh: I think the easiest is to just upload the file via `POST /web`. This is the command I use:

```
$ curl -vvv -X POST -F 'image=@data/web/index.html' http://milight-hub/web
```

sidoh:
Thanks again!

Issue #100, opened 2017-07-17
rgbw: No way to get back to white in HASS


r4nd0mbr1ck:
Not sure if I am missing something but using the HASS UI, there doesn't seem to be a way to switch a RGBW bulb back to white when it is currently in colour. 

I am thinking one way to achieve this maybe to hijack the color temperature setting (or white value?) and force any color temperature command to switch to white mode?

Note: this is using mqtt_json

sidoh:
Ah yeah, I think someone else brought this up in the comments of another ticket. Definitely not ideal. Thanks for raising.

The UDP integration handles this by special-casing the user selecting a color near white. This is not exactly the greatest in terms of user experience, but it feels like the "honest" way to do it.

r4nd0mbr1ck:
Can you point me to where in the code the UDP handles the near white? I couldn't figure it out.

How about using the effect payload from mqtt_json. It is a bit clunky because the user will have to configure the effects list exactly (say "set_white"). On the other hand, handling the effects payload opens up options for things like night_mode, disco_mode, etc. from within HASS.

sidoh:
It's [here](https://github.com/happyleavesaoc/python-limitlessled/blob/master/limitlessled/group/rgbw.py#L47), I think:

```python
    def color(self, color):
        """ Set group color.
        Color is set on a best-effort basis.
        :param color: RGB color tuple.
        """
        if color == RGB_WHITE:
            self.white()
            return
        self._color = color
        self.hue = hue_of_color(color)
```

Hijacking effects is interesting. Still need to play around with those in HASS. I'll take a look sometime soon.

bhaap:
Or just when you get a white value in the mqtt message you set the bulb into white mode? you will only get the white value in hass if you choose to change the white value slider.

This is the only thing thats holding me from switching from the official milight gateways to this instead.

sidoh:
Should just be adding some logic here:

https://github.com/sidoh/esp8266_milight_hub/blob/master/lib/MiLight/MiLightClient.cpp#L264

Might be best to handle regardless of whether color is set via rgb or hue/sat, but not too tricky. 

I'm away for a few weeks, but if you're interested in contributing, happy to look at a PR in the meantime.

sidoh:
Fix added in v1.5.0 thanks to @r4nd0mbr1ck!

PR #106, opened 2017-08-24
Switch to white if white_value in payload

r4nd0mbr1ck:
I have added code to switch to white mode if the payload contains white_value

This is for Issue #100

### DIFF
--- a/lib/MiLight/MiLightClient.cpp
+++ b/lib/MiLight/MiLightClient.cpp
@@ -267,16 +267,23 @@ void MiLightClient::update(const JsonObject& request) {
     uint8_t r = color["r"];
     uint8_t g = color["g"];
     uint8_t b = color["b"];
+    //If close to white
+    if( r > 256 - RGB_BOUNDARY && g > 256 - RGB_BOUNDARY && b > 256 - RGB_BOUNDARY)
+    {
+        this->updateColorWhite();
+    }
+    else
+    {
+      double hsv[3];
+      RGBConverter converter;
+      converter.rgbToHsv(r, g, b, hsv);
 
-    double hsv[3];
-    RGBConverter converter;
-    converter.rgbToHsv(r, g, b, hsv);
+      uint16_t hue = round(hsv[0]*360);
+      uint8_t saturation = round(hsv[1]*100);
 
-    uint16_t hue = round(hsv[0]*360);
-    uint8_t saturation = round(hsv[1]*100);
-
-    this->updateHue(hue);
-    this->updateSaturation(saturation);
+      this->updateHue(hue);
+      this->updateSaturation(saturation);
+    }
   }
 
   if (request.containsKey("level")) {
@@ -298,11 +305,6 @@ void MiLightClient::update(const JsonObject& request) {
     );
   }
 
-  //Homeassistant - Switch to white if white value is set
-  if (request.containsKey("white_value")) {
-    this->updateColorWhite();
-  }
-  
   if (request.containsKey("mode")) {
     this->updateMode(request["mode"]);
   }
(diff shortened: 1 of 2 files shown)


sidoh:
Awesome, thanks for taking a look at this.

`white_value` is a metric, not a boolean, right? The semantics are a little unclear, but I think it amounts to `brightness` restricted to dedicated white LEDs:

https://home-assistant.io/components/light/

I think other HASS milight integrations (and perhaps all other colored bulbs in HASS) switch to white mode when a color close to white on the color picker is selected. Is that not the case in your experience?

r4nd0mbr1ck:
Yep, white_value is a metric. 

I only have LIFX bulbs - which switch to white when I change the color temperature (they have tunable whites). I will test this with LIFX a bit more later tonight but I haven't yet noticed them switching to white via the color picker.

r4nd0mbr1ck:
Did a bit of investigating in the HA codebase. Looks like different light platforms do different things

1) LimitlessLED - uses a white 'effect' and also using a color close to white min(color) > 256 - boundary. So two ways of switching to white. I am just worried that using the color picker may not be accurate enough.

2) LifX - switches to white when color_temp is set. This wouldn't work for the RGBW bulbs as they don't support color temp.

I haven't managed to find how this works for others - lightify, tplink, etc. They seem to be native to the bulbs and/or the python libraries being used. 

P.S.: I have also written code to use the "effect" to set night_mode on the CCT bulbs. I might send that through a PR later.  Perhaps I should include set_white also as an effect irrespective of how we go with implementing the current issue.

r4nd0mbr1ck:
I think I agree that using the RGB would be a better way to go rather than introduce a parameter that doesn't serve it's origibal intent. Will try and work on that.

sidoh:
The effect route sounds great too, especially if it's something the existing LimitlessLED integration already does. 

Btw - changing color temp for milight will also switch RGB+CCT bulbs to white mode.

Thanks again for looking into this :)

lib/MiLight/MiLightClient.h
```diff
@@ -11,7 +11,8 @@
 //#define DEBUG_PRINTF
 
 #define MILIGHT_DEFAULT_RESEND_COUNT 10
-
+//Used to determine close to white
+#define RGB_BOUNDARY 40
```
sidoh: Not a big deal, but could we call this `RGB_WHITE_BOUNDARY` or something along those lines?

lib/MiLight/MiLightClient.cpp
```diff
@@ -267,16 +267,23 @@ void MiLightClient::update(const JsonObject& request) {
     uint8_t r = color["r"];
     uint8_t g = color["g"];
     uint8_t b = color["b"];
+    //If close to white
+    if( r > 256 - RGB_BOUNDARY && g > 256 - RGB_BOUNDARY && b > 256 - RGB_BOUNDARY)
```
sidoh: Would you mind updating the styling? e.g.:

```
if (...) {
  ...
} else {
  ...
}
```

to be consistent with the rest of the project.

sidoh:
Looks good! Thanks for taking care of this. Were you able to test out the changes?

r4nd0mbr1ck:
I have pushed two commits through - one to fix the code styles and the other to handle effects. 

I have tested the white effect against a RGBW bulb and that appears to work. However, night mode doesn't seem to work. Does the RGBW bulb have a night mode?

Night mode works with the CCT bulb.

sidoh:
Looks great to me. Thanks again!

sidoh:
This is in [1.5.0-dev3](https://github.com/sidoh/esp8266_milight_hub/releases/tag/1.5.0-dev3) now. :)

PR #116, opened 2017-09-23
Add support for FUT089

sidoh:
Remaining tasks:

- [x] Fix mode support. Weird global statefulness in firmware should be in the UI instead.
- [ ] Test that LT8900 radio still works after changes due to refactoring.
- [x] Write RF packet parser.


sidoh:
Will try to test on LT8900 radio before releasing 1.6, but merging this for now.

Issue #123, opened 2017-10-05
Add a password to the WiFi in Access Point Mode


speendo:
This project uses [WiFiManager](https://github.com/tzapu/WiFiManager) to configure the WiFi settings of the ESP8266.

This means, whenever the ESP8266 can not access another WiFi it turns to Access Point Mode and provides a web page to connect to another already existing WiFi.

The code for that is [here](https://github.com/sidoh/esp8266_milight_hub/blob/master/src/main.cpp#L153):

    wifiManager.autoConnect();

The [documentation](https://github.com/tzapu/WiFiManager#documentation) of WiFiManager states that

> You can and should password protect the configuration access point.

And for a good reason: entering my *secret* WiFi passphrase to an unencrypted web page on an unencrypted WiFi is a severe security flaw. If my neighbor is sniffing my WiFi traffic, she can then extract my WiFi passphrase, enter my WiFi (and then maybe play around with my NAS). So altoghether this is not the same but right next to not having a protected WiFi altogether.

The easy solution for this problem would be something like

    wifiManager.autoConnect("ESP" + String(ESP.getChipId()), "<some password, at least 8 chars>");

Instead of just `wifiManager.autoConnect();`

(the AP name part is "stolen" from [here](https://github.com/tzapu/WiFiManager/blob/master/WiFiManager.cpp#L135)).

sidoh:
Thanks for raising this.  It's something I've been meaning to do.  I actually just configure my ESPs with wifi credentials over serial for this reason.

I'll work this into v1.6.

Issue #120, opened 2017-10-04
Should "Clear Wifi Config" bring WIFIManager back?


speendo:
Hello!

I have my esp8266 milight hub working and connected to a wifi.

Now I want to connect it to a different wifi, while the old wifi is still present.

I thought the button "Clear Wifi Config" in the web interface would clear the old wifi settings so that I could enter new settings, once the esp8266 has rebooted.

Apparently this is not the case for me. When I press the "Clear Wifi Config" button, the esp seems to reboot, but it does not go to Access Point Mode. Instead it seems to reconnect to the old wifi.

Is this happening on purpose? Otherwise there might be a bug.

sidoh:
Yes, it's supposed to.  Looks like it's a common problem with the EPS8266 SDK:

https://github.com/esp8266/Arduino/issues/1494

Sounds like the SDK puts wifi credentials from RAM into flash when restarting, which completely undoes `ESP.eraseConfig()`.

I probably won't have a chance to look into this for a little while, but if you feel like looking into it yourself, the relevant code is [here](https://github.com/sidoh/esp8266_milight_hub/blob/master/lib/WebServer/MiLightHttpServer.cpp#L109) (search for `clear_wifi_config`).

Some things to try:

1. Use `ESP.reset()` instead of `ESP.restart()`
2. Force a crash ("best" way is probably by triggering a watchdog reset with `while(true);`)

If you're just looking for a quick solution, clearing flash with esptool.py will do the trick:

```
$ esptool.py --port /dev/cu.SLAB_USBtoUART erase_flash
```

speendo:
Thank you, sidoh!

I have to state that I have never done any ESP programming yet (some Arduino, though).

However, I looked at this issue and didn't really appreciate the two solutions you stated so much. It seemed to me, both of them are more hacks (provoking some behavior that would lead to the ESP forgetting the WiFi as a side effect) than actual solutions.

So I digged deep into the [ESP Arduino Core](https://github.com/esp8266/Arduino) and found the [disconnect method](https://arduino-esp8266.readthedocs.io/en/latest/esp8266wifi/station-class.html#disconnect) in the STA class.

So I added

    #include <ESP8266WiFi.h>

to the head of `MiLightHttpServer.cpp` and rewrote line 114 to this

    //ESP.eraseConfig();
    WiFi.disconnect(false);

It seems like this did the trick.

I can provide a fork for your project tomorrow (today I have go to sleep, I'm tired ;-)).

[...]

PR #125, opened 2017-10-06
Solving issues #120 and #123

speendo:
As promised, this pull request solves #120 and also #123.

I also changed `README.md` and added the wifi password. However, you possibly might want to put this information at a more prominent place.

Hope that helps :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,8 @@ This project uses [WiFiManager](https://github.com/tzapu/WiFiManager) to avoid t
 
 When the ESP powers on, you should be able to see a network named "ESPXXXXX", with XXXXX being an identifier for your ESP. Connect to this AP and a window should pop up prompting you to enter WiFi credentials.
 
+The network password is "**milightHub**".
+
 #### Get IP Address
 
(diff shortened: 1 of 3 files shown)


sidoh:
Looks awesome.  thank you again for looking into this.  I'll take a closer look later today or tomorrow.

speendo:
I'm glad if I can help.
Please be especially careful about [this line](https://github.com/speendo/esp8266_milight_hub/blob/b046fcf9472ffef4116280e63be130a4500e671d/lib/WebServer/MiLightHttpServer.cpp#L115). I didn't look close enough at your project to understand, if there was another way to retrieve a WiFiManager object.

Probably it doesn't matter after all. However, to use an already existing object would be nicer from... I don't know... maybe memory perspective?

sidoh:
I don't think there's a nice way to retrieve the WiFiManager object used in main.cpp.  Would either need to be injected into the HTTP server, or yanked in via `extern`, neither of which I'm crazy about.  Don't think there's any harm putting a new one on the stack.

btw - I'm going to point this at v1.6.0 instead of master.

PR #139, opened 2017-10-20
Fixed night-mode for RGBW-Bulbs

### LINKED ISSUE #136: Nightmode does not work
Using some RGBW-Bulbs (FUT016) with my own switching-app (with the milight-php-api). Nightmode does work as expected with the original bridges. But using this milight hub, the bulbs just turns off.
Tried latest v1.6-dev-release and the 1.5.0 stable.

Any idea?

bombcheck:
My attempt to fix the broken night mode for rgbw bulbs. Added missing command- and remote-button-codes and changed the necessary procedures.

Fixes #136


lib/MiLight/RgbwPacketFormatter.cpp
```diff
@@ -86,10 +86,23 @@ void RgbwPacketFormatter::updateColorWhite() {
 }
 
 void RgbwPacketFormatter::enableNightMode() {
-  uint8_t button = STATUS_COMMAND(ON, groupId);
-
-  command(button, 0);
-  command(button | 0x10, 0);
```
sidoh: I think perhaps what was there before almost gives you what you want.  It looks like the commands should actually be `STATUS_COMMAND(OFF, groupId) | 0x10`.  This would avoid needing to case each of the groups, and needing separate command IDs for night mode.

lib/Udp/V5MiLightUdpServer.cpp
```diff
@@ -64,6 +81,51 @@ void V5MiLightUdpServer::handleCommand(uint8_t command, uint8_t commandArg) {
         );
         break;
 
+      case UDP_RGBW_GROUP_ALL_NIGHT:
```
sidoh: Would we want to `client->prepare(&FUT096Config, deviceId, 0)` here (and do the same thing for the other groups in the other cases)?  Otherwise we'll send the white mode command to the last group a packet was received for.

lib/Udp/V5MiLightUdpServer.cpp
```diff
@@ -19,12 +19,29 @@ void V5MiLightUdpServer::handleCommand(uint8_t command, uint8_t commandArg) {
...
+
     this->lastGroup = groupId;
   // Command set_white for RGBW
-  } else if (command >= UDP_RGBW_GROUP_ALL_WHITE && command <= UDP_RGBW_GROUP_4_WHITE) {
+} else if (command == UDP_RGBW_GROUP_ALL_WHITE || command == UDP_RGBW_GROUP_1_WHITE || command == UDP_RGBW_GROUP_2_WHITE || command == UDP_RGBW_GROUP_3_WHITE || command == UDP_RGBW_GROUP_4_WHITE) {
     const uint8_t groupId = (command - UDP_RGBW_GROUP_ALL_WHITE)/2;
     client->prepare(&FUT096Config, deviceId, groupId);
     client->updateColorWhite();
+
+    // Just some debugging stuff
```
sidoh: Maybe good to either remove the debugging or:

* Wrap it in the appropriate compiler flags (`DEBUG_PRINTF` for RF stuff, `MILIGHT_UDP_DEBUG` for UDP stuff).
* Use the `F` macro for string constants.  Think I saw you using this in one of your early revisions.

lib/MiLight/RgbwPacketFormatter.cpp
bombcheck: Hmmm good idea... Will give this a try later...

platformio.ini
```diff
@@ -13,7 +13,7 @@ board_f_cpu = 160000000L
 lib_deps_builtin =
   SPI
 lib_deps_external =
-  sidoh/RF24
+  sidoh/RF24#remove_progmem
```
sidoh: Should probably remove this.  Will definitely want to figure out what's going on with Windows builds for the v1.6.0 release, but hoping this won't be the fix.

lib/MiLight/RgbwPacketFormatter.cpp
bombcheck: OK... This one works quite well:

uint8_t button = STATUS_COMMAND(OFF, groupId);
command(button | 0x10, 0);

Same as with my solution: Works when using udp gateway and REST API, but must be fired twice when using the REST API and the bulb is on.

platformio.ini
bombcheck: Yes, maybe a good ideal to NOT have this in your branch. But I needed this. :)
Maybe I find some time to set up the build environment on my MacBook...

lib/Udp/V5MiLightUdpServer.cpp
bombcheck: Saw the macro but did not get its intention. Removed the stuff. It's not that important. Just using this at the beginning to get a feeling how the code works. :)

lib/Udp/V5MiLightUdpServer.cpp
bombcheck: Can you explain this?

sidoh: `client->enableNightMode()` will enable night mode for the group seen in the last command.  So if you turned on group 2, and then enabled night mode for group 1, it'd enable night mode for group 2.

The fix would be to add the `client->prepare` statement for the appropriate group.  This call tells the client to format and send packets for the specified radio, device, and group combo.

lib/Udp/V5MiLightUdpServer.cpp
sidoh: `F` puts the string in flash memory so that they don't occupy RAM.  Not super important if the statements are opt-in via a compiler flag.

lib/Udp/V5MiLightUdpServer.cpp
bombcheck: Ah I see... OK, thats why my PHP API always sends OFF before engaging night mode. :)
Worked our a fix for that based on your idea.
https://github.com/bombcheck/esp8266_milight_hub/commit/e3f3ca993036d486f42061fdfd0fc25b00aaea02

sidoh:
@bombcheck this looks like it's about ready to merge.  Think we'll just need to remove the change in `platformio.ini`.

bombcheck:
Great. Should I remove it?

sidoh:
That's probably best.  Saves the commit undoing it when merged :)

bombcheck:
OK, done. :)

sidoh:
Thanks again for fixing this!

Issue #103, opened 2017-07-25
level and brightness in MQTT publish messages


coehn1:
Hi,

First off, really great job!!!
I've got a question. When I change the brightness of a bulb, the MQTT published message contains the 'brightness  value, 0-255. And the 'state', ON or OFF. Is it possible to also send the 'level', 0-100 and the 'status' values? So the normal commands AND the redundant HomeAssistant's?

Kind regards,

Niels Coehorst
The Netherlands

sidoh:
Certainly possible. I was thinking it made sense to optimize more for the size of the messages. Probably not too hard to have something like a format string with sane defaults.

coehn1:
Sorry, I don't understand you.

Verstuurd vanaf mijn iPhone

Op 25 jul. 2017 om 19:57 heeft Chris Mullins <notifications@github.com> het volgende geschreven:



Certainly possible. I was thinking it made sense to optimize more for the size of the messages. Probably not too hard to have something like a format string with sane defaults.

sidoh:
Heh, sorry.

Guess I'm saying I'd prefer to not send all variants of each setting (brightness + level, status + state, hue + color, etc.) because the MQTT messages would start getting pretty big. 

I was proposing exposing a message format setting, something like:

```
{"state":"{{state}}", "level":"{{level}}"}
```

which you could fiddle with to get messages in the format you like.

coehn1:
Hi Chris,


Thanks, great idea!!

Verstuurd vanaf mijn iPhone

Op 25 jul. 2017 om 20:25 heeft Chris Mullins <notifications@github.com> het volgende geschreven:



Heh, sorry.

Guess I'm saying I'd prefer to not send all variants of each setting (brightness + level, status + state, hue + color, etc.) because the MQTT messages would start getting pretty big.

I was proposing exposing a message format setting, something like:

{"state":"{{state}}", "level":"{{level}}"} 

which you could fiddle with to get messages in the format you like.

sidoh:
Another potential approach: list the keys you want to include in the state payload.  Might default to something like `state, brightness, color, color_temp, bulb_mode, mode`, but you could swap out `level` for `brightness` if you wanted.

This is less flexible, but I think would perform better, and is probably less confusing. 

@Skinah does this seem like it'd do the trick for you?

Skinah:
[...]

PR #144, opened 2017-11-02
v1.6.0 release

sidoh:
## Release checklist

* [x] ( #103 ) Add ability to select which keys are sent in state updates.  Will partially address #149.
* [x] Look into merging state updates into one MQTT message.  Some discussion in #149.
* [x] Add documentation for group_state_fields in README.

#### New device support

* FUT089/B8 remotes, which allow for control of up to eight groups with one remote.

#### State tracking!

* State for all devices -- that is, *(device type, device id, group)* triplets -- is now passively tracked and persisted.
* Add a REST route to retrieve the state of a given device.
* Add support for publishing full device state to MQTT.
* Add support for sending raw commands and arguments in MQTT and REST integrations.
* REST endpoint to change bulb state now returns the new state of the bulb instead of just `true`.
* Switching devices in the UI will update settings to known values for that ID (e.g., on/off switch, brightness, etc. will be set to last known values).
* Add the ability to select which keys (`brightness`, `color`, etc.) are sent in MQTT updates (and REST state responses).  Defaults to keys used by HomeAssistant's `mqtt_json` component, but you can now select alternatives like `level`, or `hue`/`saturation` instead of `color`.

#### nRF Radio Optimizations

* @khmann cut a lot of unnecessary work and added optimizations in the nRF radio handler code which improved performance by 40% (!).

#### MQTT improvements

* MQTT state updates are published as retained messages. New subscribers to state topics will receive the last published message. This means, for example, HomeAssistant will boot with the correct bulb states.
* ( #103 ) Add ability to select which fields (`brightness`, `level`, etc.) are sent in MQTT updates and state messages.

#### Security

* Compile against Arduino 2.4.0-rc2, which includes a fix for the KRACK vulnerability in WPA2.
* @speendo added a password for the setup AP.  The password is `milightHub`.  This prevents your wifi credentials from being transmitted in plaintext.

## Bug fixes

* @bombcheck fixed a bug with handling night mode for RGBW bulbs.
* FUT092 (RGB+CCT) packets encoded with the key `0xD4` were getting decoded incorrectly.  This has been fixed.
* Removed improper usage of PROGMEM which might've caused crashes when compiled in some environments.
* Fixed crashes when compiled in a Windows environment.
* Fix edge case where wrong packets being sent in V5 UDP integrations.
* The `GET /settings` endpoint returned the raw settings file on flash, which meant newly added setting keys weren't populated with their default value.  The value returned by this endpoint is now populated with default values for otherwise undefined settings.
* Loading settings from the UI (and the `POST /settings` REST endpoint) were not working properly. This has been fixed.
* Fixed bug where packet repeats weren't set correctly when settings were saved.
* @speendo fixed a bug causing the "Clear Wifi Config" feature to not work.


bombcheck:
Merged your latest changes into my dev-branch. Works great now!

bombcheck:
PlatformIO-Beta? What happens if I do not use the beta? It compiles and runs fine with the "normal" PlatformIO-Release.

sidoh:
I've not found a good reason to not use the beta.  It's pretty stable, and installing it is easy.

If you prefer to not use the beta, you can revert https://github.com/sidoh/esp8266_milight_hub/pull/144/commits/c6bd8700b5545c0f4abee7ac148a5d5f0e8a6ae3 and install the platform manually.

bombcheck:
Ack. Just wondering why it compiles flawlessly. Using 3.5.0b2. I think this is the beta... :)

sidoh:
Yep, 3.5 is the beta. :)

bombcheck:
OK...
Just tried the REST API. Works great but I always get an '{}' as response. Same if I try to GET the status of a bulb.

bombcheck:
Any idea why I always get an empty response when using the REST API? Saved states seem to be empty, too as I get the same empty response when GETting the state via the API.

sidoh:
It will serve empty state for bulbs that haven't had any state updates.  Which commands are you issuing to the bulbs?  Could you provide some curl commands that I could try to reproduce the problem with?

sidoh:
Seems to work fine:

```
$ curl 10.133.8.158/gateways/5/rgbw/1
{}%

$ curl -X PUT -H'Content-Type: application/json' -d '{status:"ON"}' 10.133.8.158/gateways/5/rgbw/1
{"state":"ON"}%

$ curl 10.133.8.158/gateways/5/rgbw/1
{"state":"ON"}%

$ curl -X PUT -H'Content-Type: application/json' -d '{"color":{"r":0,"g":255,"b":255},"brightness":100}' 10.133.8.158/gateways/5/rgbw/1
{"state":"ON","brightness":102,"bulb_mode":"color","hue":181}%

$ curl 10.133.8.158/gateways/5/rgbw/1
{"state":"ON","brightness":102,"bulb_mode":"color","hue":181}%
```

bombcheck:
```
curl -X PUT -H 'Content-Type: application/json' -d '{status:"ON"}' 192.168.1.147/gateways/0xf14a/rgbw/1
{}

curl 192.168.1.147/gateways/0xf14a/rgbw/1
{}
```

If it works with your build then it must have to do with my one (merged your changes into my fork lately). Or maybe I should try a clean flash?

sidoh:
Yes, would recommend a clean flash and trying with one of the pre-compiled images:

```
$ curl -X PUT -H 'Content-Type: application/json' -d '{status:"ON"}' 10.133.8.158/gateways/0xf14a/rgbw/1
{"state":"ON"}%

$ curl 10.133.8.158/gateways/0xf14a/rgbw/1
{"state":"ON"}%
```

bombcheck:
Tried a clean flash. Noticed that I can not import settings file. It says "success" but does not import anything actually. Problem with states persists: Still getting only an empty response.
Did not yet tried your image (should work as you testet it) but my compiled one. Maybe a new sort of compiler issue? The progmem-bug has been fixed I think?

sidoh:
Can you try compiling from my branch?  Not sure what changes you have locally. 

It's super hard for me to debug and support a Windows dev environment because I don't have one to test on.  Any support on the debugging effort would be much appreciated.

bombcheck:
Sure. Your branch compiles without errors and runs fine. BUT I get exactly the same empty state responses.
Will try to set up the build environment on my Mac later and report back then.

sidoh:
Sounds good.

My best guess is that the Windows compiler is doing something different with this:

https://github.com/sidoh/esp8266_milight_hub/blob/v1.6.0/lib/MiLightState/GroupState.h#L93

using unions in this way is creepy, but I'm operating under the assumption that it shouldn't matter as long as the data aren't used cross-platform.  I would think it should behave consistently.

If you're feeling brave, you might try adding something like this in `main.cpp:setup()`:

```c++
GroupState testState;
printf("kelvin set: %d, kelvin value: %d\n", testState.isSetKelvin(), testState.getKelvin());
testState.setKelvin(100);
printf("kelvin set: %d, kelvin value: %d\n", testState.isSetKelvin(), testState.getKelvin());
```

bombcheck:
Build the dev-branch of my fork on my MacBook: Same issues with empty states.
Compiled your v1.6.0-branch: Same issue: Empty states.
I did not yet tried a clean flash with those new builds...

But maybe there is another strange thing going on? Any "special" tweaks in your build environment?

sidoh:
Can you try with a pre-compiled binary?

https://github.com/sidoh/esp8266_milight_hub/releases/tag/1.6.0-rc1

These are built using a totally clean-slate build environment (a container).  The tests I were running earlier were from this.

bombcheck:
Just tried: Same issue.

sidoh:
That's the code I'm running too, so I think it's gotta be something weird with your ESP.  Maybe try clearing flash with esptool and re-flashing.  Also worth looking at serial logs (maybe enable `STATE_DEBUG`).

bombcheck:
Just tried a clean flash and immediately using REST API after flashing: Same problems: Empty states and I can not import a settings-file (says "success" but does not really import anything).
The commands are executed and send correctly to the bulbs, but not saved. :(

bombcheck:
Really strange... Importing of a config file via the web gui does not work. But if I put the content of the file into a CURL request, it DOES work.

`curl -v -X PUT -H 'Content-Type: application/json' -d '#CONTENT OF SETTINGS-FILE#' milight-hub.local/settings`

sidoh:
my settings are mostly the defaults.  Kind of doubt this is the issue:

```
{
  "admin_username": "",
  "admin_password": "",
  "ce_pin": 16,
  "csn_pin": 15,
  "reset_pin": 0,
  "radio_interface_type": "nRF24",
  "packet_repeats": 50,
  "http_repeat_factor": 1,
  "auto_restart_period": 0,
  "mqtt_server": "mymqtt",
  "mqtt_username": "user",
  "mqtt_password": "hunter2",
  "mqtt_topic_pattern": "milight/:device_id/:device_type/:group_id",
  "mqtt_update_topic_pattern": "",
  "mqtt_state_topic_pattern": "milight/states/:device_id/:device_type/:group_id",
  "discovery_port": 0,
  "listen_repeats": 3,
  "state_flush_interval": 10000,
  "mqtt_state_rate_limit": 500,
  "packet_repeat_throttle_sensitivity": 0,
  "packet_repeat_throttle_threshold": 200,
  "packet_repeat_minimum": 3,
  "device_ids": []
}
```

I'll take a look at the settings.  I think Arduino 2.4.0-rc2 changed some behavior that requires the content type be set.  The UI is probably not setting it.

bombcheck:
OK, looking forward to have the settings import fixed.
I see you have configured a mqtt broker, which I do NOT have. Maybe this is the problem?

sidoh:
I doubt it's because MQTT is configured.  State management is out of band from MQTT.  I can try running on a different ESP without MQTT configured to confirm, though.

bombcheck:
OK, would be great to clarify on that. Getting a second esp before the weekend. So I do not always have to crawl into my sideboard to flash my "prod" esp (and getting my wife upset when the light control is broken during testing). :)

bombcheck:
Think we have to go for a AJAX-Request or something to set the content type when uploading the settings file via the gui...

sidoh:
lol nevermind, this is totally the issue.

```c++

  if (mqttClient) {
    GroupState& groupState = stateStore->get(bulbId);
    groupState.patch(result);
    stateStore->set(bulbId, groupState);

    // Sends the state delta derived from the raw packet
    char output[200];
    result.printTo(output);
    mqttClient->sendUpdate(remoteConfig, bulbId.deviceId, bulbId.groupId, output);

    // Sends the entire state
    bulbStateUpdater->enqueueUpdate(bulbId, groupState);
  }
```

probably just need to yank the group state update out of the conditional.  i'll test that later.

sorry about that.  good find!

bombcheck:
LOL. Looking forward to nail down that greedy glitch.

sidoh:
Should be fixed in https://github.com/sidoh/esp8266_milight_hub/pull/144/commits/63e4acaccabc3592d1d406fd2f52059fd95ead97.  

Will take a look at the settings thing.  

Also noticing RGBW/CCT packets are misclassified when picked up on adjacent channels, so will add a check for that as well.

bombcheck:
Fix confirmed. :)

bombcheck:
Noticed that setting night mode (rgbw) is not reflected in the state. bulb_mode stays at the last mode ("white" / "color" and state is OFF.

sidoh:
Squashed some bugs in [v1.6.0-rc2](https://github.com/sidoh/esp8266_milight_hub/releases/tag/1.6.0-rc2).

Added a checklist with TODOs before release.

PR #182, opened 2017-12-10
Bufix checksum in CCT packet.

Erriez:
Add checksum of the 7 packet bytes (including packet length = 7).
Without it, several CCT bulbs cannot be linked and do not respond
on commands. The CctPacketFormatter.cpp protocol is now identical
to the FUT006 remote CCT controller.

The CCT protocol is as follows:
```
  Byte 0: Packet length = 7 bytes
  Byte 1: CCT protocol 0x5A
  Byte 2 and 3: Device ID (16-bit)
  Byte 4: Zone 0..4
  Byte 5: Bulb command
  Byte 6: Packet sequence number 0..255
  Byte 7: Checksum over previous bytes 0..6
  Byte 8: CRC LSB over bytes 0..7
  Byte 9: CRC MSB over bytes 0..7
```

Tested with FUT011: 5W GU10 Dual White LED Spotlight, MiLight Hub web interface in CCT mode.

### DIFF
--- a/lib/MiLight/CctPacketFormatter.cpp
+++ b/lib/MiLight/CctPacketFormatter.cpp
@@ -35,6 +35,18 @@ void CctPacketFormatter::initializePacket(uint8_t* packet) {
   // Byte 9: CRC MSB
 }
 
+void CctPacketFormatter::finalizePacket(uint8_t* packet) {
+  uint8_t checksum;
+
+  // Calculate checksum over packet length .. sequenceNum
+  checksum = 7; // Packet length is not part of packet
+  for (uint8_t i = 0; i < 6; i++) {
+    checksum += currentPacket[i];
+  }
+  // Store the checksum in the sixth byte
+  currentPacket[6] = checksum;
+}
+
 void CctPacketFormatter::updateBrightness(uint8_t value) {
   valueByStepFunction(
     &PacketFormatter::increaseBrightness,
@@ -54,21 +66,11 @@ void CctPacketFormatter::updateTemperature(uint8_t value) {
 }
 
 void CctPacketFormatter::command(uint8_t command, uint8_t arg) {
-  uint8_t checksum;
-
   pushPacket();
   if (held) {
     command |= 0x80;
   }
   currentPacket[CCT_COMMAND_INDEX] = command;
-
-  // Calculate checksum over packet length .. sequenceNum
-  checksum = 7; // Packet length is not part of packet
-  for (uint8_t i = 0; i < 6; i++) {
-    checksum += currentPacket[i];
-  }
-  // Store the checksum in the sixth byte
-  currentPacket[6] = checksum;
 }
 
 void CctPacketFormatter::updateStatus(MiLightStatus status, uint8_t groupId) {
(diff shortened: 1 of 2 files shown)


lib/MiLight/CctPacketFormatter.cpp
```diff
@@ -37,11 +54,24 @@ void CctPacketFormatter::updateTemperature(uint8_t value) {
...
+
+  // Calculate checksum over packet length .. sequenceNum
+  checksum = 7; // Packet length is not part of packet
+  for (i = 0; i < 6; i++) {
+    checksum += currentPacket[i];
+  }
+  // Store the checksum in the sixth byte
+  currentPacket[6] = checksum;
+
+  Serial.println(checksum, HEX);
```
sidoh: Can we remove this println?

lib/MiLight/CctPacketFormatter.cpp
```diff
@@ -37,11 +54,24 @@ void CctPacketFormatter::updateTemperature(uint8_t value) {
...
+
   pushPacket();
   if (held) {
     command |= 0x80;
   }
   currentPacket[CCT_COMMAND_INDEX] = command;
+
+  // Calculate checksum over packet length .. sequenceNum
+  checksum = 7; // Packet length is not part of packet
+  for (i = 0; i < 6; i++) {
```
sidoh: Unless `i` is going to be used outside of the scope of the loop (doesn't look like it will), could we initialize it here?

lib/MiLight/CctPacketFormatter.cpp
```diff
@@ -37,11 +54,24 @@ void CctPacketFormatter::updateTemperature(uint8_t value) {
 }
 
 void CctPacketFormatter::command(uint8_t command, uint8_t arg) {
```
sidoh: I think it's probably fine to have it here, but there's a default-noop method in `PacketFormatter` called `finalizePacket`.  I think it'd make sense to override this method and calculate the checksum there.  [Here's an example](https://github.com/sidoh/esp8266_milight_hub/blob/master/lib/MiLight/V2PacketFormatter.cpp#L54-L56) of it being used in the 9-byte protocol formatter.

lib/MiLight/CctPacketFormatter.cpp
Erriez: I assume you prefer this code instead?

```
for (uint8_t i = 0; i < 6; i++) {
```

That's fine.

lib/MiLight/CctPacketFormatter.cpp
Erriez: Oops, that was a debug print. Good catch!

lib/MiLight/CctPacketFormatter.cpp
Erriez: Yes, your suggestion would be a better solution. I'll try this later.

Erriez:
Seems to work now. :-) 
Thanks for developing this great project!

sidoh:
Fantastic 🎉 !

Thank you very much for fixing this.  I'll tag a new release shortly.

Issue #170, opened 2017-11-25
Sporadic buffer overflow in CCT brightness commands


sidoh:
Discussion and detail in this comment thread:

http://blog.christophermullins.com/2017/02/11/milight-wifi-gateway-emulator-on-an-esp8266/comment-page-3/#comment-2659

vantees:
Hi. I can confirm that with CCT bulbs interface stops working. After sending just one command ON/OFF it does not respond any longer. You have to restart ESP. Tweaking ESP does not help. I have D1_mini and 1.6.2 (d1_mini) firmware.

Skinah:
I have the same issue with brightness and colour changing CCT globes. The behaviour is like the espmilight hub is trying to do a transistion as you see the level slowly fade instead of jumping directly to what you select. On and Off is also very very slow to respond. Also the colour temp does not get published to the mqtt state.

sidoh:
> The behaviour is like the espmilight hub is trying to do a transistion as you see the level slowly fade instead of jumping directly to what you select. On

This is an unavoidable limitation of the protocol the bulbs use.  With the CCT protocol, there is no command to set brightness or color temperature directly.  There are only increase/decrease commands.

espMH tries to expose a way to set each of these directly with some hackery.  It sends the down command 10 times, followed by the appropriate number of up commands.  This is really brittle and honestly I would not recommend using it in its current state.

A lot of newer CCT devices support some of the newer protocols (like RGB+CCT).  If that's an option for you, would definitely recommend it.

Otherwise, happy to take suggestions on how to better handle brightness and temperature on the old CCT protocol.

khmann:
It's a little hard to justify the complexity (underlying code change), but here's how you make it the best:

Right now you send 10x 30 repeats (ie: unique sequence numbers) warm/cold packets to get to a known state.  This is a little choppy and could be inaccurate.  You can save time and get better guarantee by sending 256x unique sequence (don't repeat) warm/cold packets It's enough to ensure the bulb gets where you want it. Save a little time by ramping full warm/cold whichever is closest to target temperature.  Then adjust as normal.

Also, don't those bulbs have a "full on"?  what if you always ramp the bulb down when asked to turn it off via MQTT.  Then when turning on, either "resume" (=lowest) or full on and adjust from there.

The same (always use unique sequence) applies to old RGB protocol brightness as well

sidoh:
[...]

PR #193, opened 2018-01-02
Modified to work with Adafruit Feather Huzzah

hastarin:
This turned out to be exceptionally easy.  Thanks for all your work here!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /web/node_modules
 /web/build
 /dist/*.bin
+.vscode/
--- a/lib/Settings/Settings.h
+++ b/lib/Settings/Settings.h
@@ -71,8 +71,8 @@ public:
     adminUsername(""),
     adminPassword(""),
     // CE and CSN pins from nrf24l01
-    cePin(D0),
-    csnPin(D8),
+    cePin(16),
+    csnPin(15),
     resetPin(0),
     radioInterfaceType(nRF24),
     deviceIds(NULL),
--- a/platformio.ini
+++ b/platformio.ini
@@ -71,3 +71,13 @@ extra_scripts = ${common.extra_scripts}
 lib_deps =
   ${common.lib_deps_builtin}
   ${common.lib_deps_external}
+
+[env:huzzah]
+platform = ${common.platform}
+framework = ${common.framework}
+board = huzzah
+build_flags = ${common.build_flags} -Wl,-Tesp8266.flash.1m64.ld -D FIRMWARE_VARIANT=huzzah
+extra_scripts = ${common.extra_scripts}
+lib_deps =
+  ${common.lib_deps_builtin}
+  ${common.lib_deps_external}
--- a/platformio.ini
+++ b/platformio.ini
@@ -76,7 +76,7 @@ lib_deps =
 platform = ${common.platform}
 framework = ${common.framework}
 board = huzzah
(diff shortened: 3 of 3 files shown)


hastarin:
Actually I'll need to do more work it seems.

Things work fine to start but then I just get an error until I reset the board.

> ERROR: packet buffer full!  Cannot buffer a new packet.  THIS IS A BUG!

hastarin:
So it turns out that is an existing issue.  #170 

I'll have to see if I can find the time to investigate if there is anything I can do to help out there.

sidoh:
Is this with the "CCT" remote type?  I think we just need to increase the size of the buffer here:

https://github.com/sidoh/esp8266_milight_hub/blob/master/lib/MiLight/PacketFormatter.h#L12-L15

Would be nice to know why 140 bytes is wrong, though.  I can't think of a reason more than 140 bytes would be necessary.  Could just be a silly off-by-one error in the bounds checking:

https://github.com/sidoh/esp8266_milight_hub/blob/master/lib/MiLight/PacketFormatter.cpp#L119-L123

Anyway, PR looks great, and as long as you're using the CCT remote type, I think the error you're seeing is unrelated.

hastarin:
@sidoh Yes this is with the CCT remote type.  I've just confirmed it's with the brightness command.

I'll look into it today.  

From my days of controlling these things via Tasker and my UDP Sender app I know the UDP commands for full brightness were to send the turn on twice with 120ms delay between them.

So, aside from the "ouch it's too bright" factor, one option may be to send the command to go full brightness, if I can figure out how to do that, and then ramp down.

hastarin:
I've confirmed the Color Temperature causes this too.   I'm currently just experimenting with the built in UI and trying to relearn C++ since it's been 20 years since I touched it.

I noticed the light seems to go through 2 ramps of down/up instead of one.  So I've currently just doubled the buffer size to see if that helps and been unable to trigger the bug since.

Now to try and figure out why the double ramp, and what I can do to prevent it, or change the behavior to be a bit smarter about moving states.

I do wonder if having a "resync" option would be of more use than trying to do this on every change?

sidoh:
Thanks for digging into that!  Odd that it's doing two cycles -- that certainly shouldn't be happening.  I'll drop some notes in the the issue you found (#170).  As this PR is unrelated, let's continue the discussion on this bug there.

A thought on the PR -- is 1MB flash (with 64KB SPIFFS) enough to support OTA updates?  If not, we should probably document that somehow.

hastarin:
I honestly have no idea on the OTA updates.  From what I can tell the board I'm using has 4MB flash:
https://learn.adafruit.com/adafruit-feather-huzzah-esp8266

sidoh:
Ah ok.  Asking because of the build flag:

```
-Wl,-Tesp8266.flash.1m64.ld
```

Which assumes the chip has 1MB of flash.  Does it work without this flag?

hastarin:
Sorry that's me copy/pasting and not knowing what I'm doing.

Should I be using the following instead?
`
-Wl,-Tesp8266.flash.4m1m.ld
`

Or just remove it entirely?

It does work without it.  I lost my settings, but I guess they were stored at a different offset.  Saving things again still survived a reset.

sidoh:
Gotcha.  Probably best to remove it if it works.  Think it gets auto-assigned without it, and it's usually the right thing.

hastarin:
All done.

sidoh:
Awesome.  Thanks, @hastarin!

PR #196, opened 2018-01-04
Default color temp rgb_cct white mode 100

starkillerOG:
Change the default color tempreture for the rgb_cct protecol when pressing the "white mode" button from 0 (cold white) to 100 (warm white).
In a later update this schould become the last known color temperature stored in memory from the bulb state.


sidoh:
:tu:

Thanks @starkillerOG!

PR #199, opened 2018-01-07
add field to better support HASS effects

bartrumb:
Can this be merged?


sidoh:
Ah, loose ends.  I forget what I was waiting for with this... think I just wanted to test it out.  But it does seem to work.

Does it work for you?

PR #216, opened 2018-02-13
Improved Windows build support and warning cleanup

chrismidgley:
I found that on Windows (using either VS Code or Atom) the PlatformIO builds were not working.  They did work from a CMD shell, but from VS Code and Atom it is PowerShell and the build would fail to generate the web page (it would always use the cache).  I made changes to .build_web.py to fix these issues (two different problems, relating to ".cmd" scripts and to a problem with NPM environment being overridden to platform making dev not work), as well as added some additional error handling.

I also cleaned up the compile time warnings (just two of them).  Worth noting, in WebServer/WebServer.h the macro HTTP_UPLOAD_BUFLEN was defined as 2048, but it never took hold because in ESP8266WebServer/src/ESP8266WebServer.h it was previously defined as 20. 

I also added some missing properties to web/package.json to remove those warnings.

The only warnings I still get are related to outdated NPM modules, but the updated modules are brand new (versions x.0.0) so it felt premature to make that shift.


sidoh:
Thanks very much for the PR!  This looks like good stuff.

It looks like there are two changes here --

* Bug fixes in the build setup
* Some changes around how commands are handled

I'm thinking it'd be best to put the firmware changes and the build changes in separate PRs.  Does that sound right to you?

chrismidgley:
Yes, that makes a lot of sense.  I should have broken it all apart, and used separate branches.  I’ve got a number of other proposed changes coming, but I need to do just that – break apart the code changes into separate branches so I can issue multiple pull requests against individual PRs.

 

Would you prefer me to remove the pull request and break it apart as well to keep it clean?

 

Thanks,

 

Chris

PR #218, opened 2018-02-15
Fixed Windows build issues

chrismidgley:
Fixed issue with building on WIndows under Atom or VS Code.  Both use PowerShell, which doesn't work correctly with CMD files, and requires backslash instead of forward.  Also improved error handling to help debug problems.


.build_web.py
```diff
@@ -20,14 +20,25 @@ def build_web():
         os.chdir("web")
         print("Attempting to build webpage...")
         try:
-            print check_output(["npm", "install"])
-            print check_output(["node_modules/.bin/gulp"])
+            if platform.system() == "Windows":
+                print check_output(["npm.cmd", "install", "--only=dev"])
+                print check_output(["node_modules\\.bin\\gulp.cmd"])
+            else:
+                print check_output(["npm.cmd", "install"])
```
sidoh: should this be `npm` instead of `npm.cmd`?

.build_web.py
```diff
@@ -20,14 +20,25 @@ def build_web():
...
+                print check_output(["node_modules/.bin/gulp"])
             copyfile("build/index.html.gz.h", "../dist/index.html.gz.h")
-
+        except OSError as e:
+            print "Encountered error OSError building webpage: ", e
+            if e.filename:
+                print "Filename is", e.filename
```
sidoh: not that it matters much, but should there be a space at the end of the string here?

chrismidgley:
Crud.  Yes, the non-Windows leg should be "npm" not "npm.cmd".  My bad, as I can't test it (right now at least) on Linux.  

The space was there before and I left it, but I'll remove it as well.

Update coming in a bit.

sidoh:
The Travis build should run automatically when you push commits to a PR.  I was just checking the build logs here:

https://travis-ci.org/sidoh/esp8266_milight_hub/builds/341688382

chrismidgley:
Pulling - will republish under "cmidgley" account

PR #219, opened 2018-02-15
Cleaned up to remove warnings

chrismidgley:
Minor fixups to remove some compile time warnings.

### DIFF
--- a/lib/ESP8266WebServer/src/ESP8266WebServer.h
+++ b/lib/ESP8266WebServer/src/ESP8266WebServer.h
@@ -129,7 +129,7 @@ template<typename T> size_t streamFile(T &file, const String& contentType){
     sendHeader("Content-Encoding", "gzip");
   }
   send(200, contentType, "");
-  return _currentClient.write(file, HTTP_DOWNLOAD_UNIT_SIZE);
+  return _currentClient.write(file);
 }
 
 protected:
--- a/lib/WebServer/WebServer.h
+++ b/lib/WebServer/WebServer.h
@@ -7,7 +7,6 @@
 #include <PatternHandler.h>
 
 #define HTTP_DOWNLOAD_UNIT_SIZE 1460
-#define HTTP_UPLOAD_BUFLEN 2048
 #define HTTP_MAX_DATA_WAIT 1000 //ms to wait for the client to send the request
 #define HTTP_MAX_POST_WAIT 1000 //ms to wait for POST data to arrive
 #define HTTP_MAX_SEND_WAIT 5000 //ms to wait for data chunk to be ACKed
--- a/web/package.json
+++ b/web/package.json
@@ -1,7 +1,11 @@
 {
   "name": "esp8266-milight-hub-web",
   "version": "1.0.0",
-  "description": "",
+  "description": "Replacement for a Milight/LimitlessLED hub hosted on an ESP8266",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/sidoh/esp8266_milight_hub.git"
+  },
   "main": "gulpfile.js",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1"


sidoh:
Thanks for fixing.

A PR I made to the Arduino SDK was merged a while back, and I don't think it's necessary to have the super gross copy/paste of the HTTP server anymore.  But no reason to have noisy compiler warnings in the meantime.

lib/ESP8266WebServer/src/ESP8266WebServer.h
```diff
@@ -129,7 +129,7 @@ template<typename T> size_t streamFile(T &file, const String& contentType){
     sendHeader("Content-Encoding", "gzip");
   }
   send(200, contentType, "");
-  return _currentClient.write(file, HTTP_DOWNLOAD_UNIT_SIZE);
+  return _currentClient.write(file);
```
sidoh: Why does this need to change?

chrismidgley: write no longer accepts the size as a parameter, it comes from the file stream directly (that version of write is depreciated)

chrismidgley:
I just attempted to merge in the latest web server (though I also was merging in an updated WifiManager on the development branch, so grain of salt on this) ... but after making some small fixes to accept the new web server it crashes with glory!  So without further debugging, I've backed that stuff out for now in my branch.  Might be worth trying again with just the latest web server?  I had to change WebServer's "server" property to be a pointer to the object rather than directly instantiated (as you can no longer do that due to a missing copy operator), and therefore added dynamic allocation (and delete on destruction) and adjusted all server.XXX to server->XXX.  That made it compile again, but as I said it crashed.  Do you have a reasonable way to debug on the ESP?  I've been resorting to old school Serial.printf - I've seen that you can do some level of crash dump but I've failed so far to get that to work...

sidoh:
The only reason I forked it is because at some point espMH was running pretty tight on free heap, and the upload buffer was statically allocated, and the size was an unconfigurable 2KB.  The maintainers of the project merged a PR I submitted to make the parameter configurable as a compiler flag, so should be able to just set that.

If you have the appetite, probably The Right™ solution is to nuke the fork of the HTTP server and use the one packaged with the SDK.  Is this what you're saying you tried?

Unfortunately I'm not familiar with a better way to debug.  I know that PlatformIO has some fancy debugging functionality, but it's not something I have any experience with.  

You can decode the stack traces with [EspExceptionDecoder](https://github.com/me-no-dev/EspExceptionDecoder).  I paste it into a file and use a command like this to decode:

```
$ java -jar /opt/EspStackTraceDecoder.jar /Users/cmullins/.platformio/packages/toolchain-xtensa/bin/xtensa-lx106-elf-addr2line .pioenvs/esp32/firmware.elf /tmp/a
```

It's not perfect, but gives you some good hints.  Sometimes the stack seems to include stuff it shouldn't.  Guessing because of weird behaviors with yielding or something like that.

chrismidgley:
I can give it a shot with The Right solution - wasn't very hard last time I tried, but I won't spend a bunch of time if it crashes and burns all over the place!

Is there still a concern around size?  I know that 512MB Flash isn't large enough (for the smaller ESP units) - was that the issue, or was it a stack space issue, or...?     If it remains an issue (please help me understand why so I avoid it!), do you know the compiler flag (or can point me to the info)?

sidoh:
The issue was RAM.  ESP8266 has 64 KB of RAM, and most of that is gobbled up by the SDK.  A bare-bones app that uses WiFi is left with somewhere between 20-30 KBs.

Sorry, should've mentioned what the flag was.  The compiler flag is `HTTP_UPLOAD_BUFLEN`.  Think I set it to 20 bytes or something.  Given that it didn't really affect upload speeds in a noticeable way, I'd rather reclaim the 1980 bytes of RAM, even if it's not immediately useful.

I think the firmware image is too large for boards using a 512KB flash chip, but that's a separate issue.

chrismidgley:
Pulling - will try to do The Right Thing.

PR #222, opened 2018-02-15
Change BulbId == to exclude device type

chrismidgley:
When changing between devices (controllers/remotes), the system uses a different set of cached properties for each device type.  However, many combinations of devices can controller the same device ID/group (aka, bulbs) and changing to a different set of cached properties results in confusion if you switch between devices.  This proposed change removes device type from the "==" operator on the BulbId class, making it so that changing between device types continues to share the same properties (such as hue and brightness).

### DIFF
--- a/lib/MiLightState/GroupState.cpp
+++ b/lib/MiLightState/GroupState.cpp
@@ -47,10 +47,15 @@ void BulbId::operator=(const BulbId &other) {
   deviceType = other.deviceType;
 }
 
+// determine if now BulbId's are the same.  This compared deviceID (the controller/remote ID) and
+// groupId (the group number on the controller, 1-4 or 1-8 depending), but ignores the deviceType
+// (type of controller/remote) as this doesn't directly affect the identity of the bulb
 bool BulbId::operator==(const BulbId &other) {
   return deviceId == other.deviceId
-    && groupId == other.groupId
-    && deviceType == other.deviceType;
+    && groupId == other.groupId;
+    // used to contain this test below for deviceType, removed as it does not identify the bulb, but instead
+    // identifies which remote is being used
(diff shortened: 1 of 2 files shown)


sidoh:
The thinking here is that some bulbs can use multiple device types, and keying on device type when storing state means there's more than than one source of truth?

On the flip side:

* not all devices do this
* I don't think any device supports all device types.  

The FUT105s I have, for example, support both FUT089 and FUT092 protocols, but:

* it "binds" to whichever it's paired with.  
* for example, if you pair with a FUT089 remote, the bulb doesn't respond to FUT092 commands from the same device ID.  
* you _can_ pair with both a FUT089 and a FUT092 having the same device ID, but you have to do so separately.

This kind of behavior is why device type is incorporated in the state key.  Seems like both behaviors exist, but I'm not sure how to resolve that.

chrismidgley:
I agree this can go either way.  My take was it felt more natural as a user to maintain state across types, because most likely that was my intent (and makes it easier when I make a mistake and choose the wrong type).  The RGB+CCT and FUT089 both control the same bulb, both bind the same - yet when I switch between them my state is incorrect. 

Honestly not a big deal either way.  Eventually the answer should be to build up a list of the various configured groups (likely with some name/tag) and track the correct type to use with them, and then the user simply selects a group and automatically the right type is selected (and it would make the REST and MQTT interfaces nicer).  I think I saw a request similar to this concept.

Speaking of, one thing that I still don't have my head around is that a single bulb can be bound to multiple groups.  That makes state inconsistent as the code doesn't understand that relationship.  Another problem for another day...

So take it or leave it on this one!  Feel free to drop it if you don't want it.

chrismidgley:
Will republish shortly

PR #223, opened 2018-02-15
Making saturation and temperature work when on inconsistent modes

chrismidgley:
Currently, on v2 devices,  it can be confusing when changing saturation while in White mode, or changing temperature while in Color mode.  The bulb devices themselves can not do this, and the result is that the UI updates are confusing and the REST API might not affect the change desired.  On the more modern bulbs, with RGB+CCT, you can have different temperature while looking at a color, even though the bulb doesn't support updating it.  

This PR adds a capabilities to both RgbCct and FUT089 to determine what mode the device is currently in (by looking at the state) and if the device is in an incompatible mode, it switches to the correct mode, makes the change, and switches back.  For example, if in COLOR and changing Temperature, it will briefly change to White, change the temperature, and then change it back to the original color.


chrismidgley:
Will republish shortly.

PR #224, opened 2018-02-15
LED status to indicate state of Wifi and messages

chrismidgley:
This PR includes a new feature to update an LED with status of the device.  See the README.md for details on the different flashing modes.  It indicates if the Wifi is in AP mode, connected or failed, as well as signals when packets are traveling.  The UI has been modified to support setting the pin of the LED, but it defaults to pin 2 (the built-in blue LED on NodeMCU) and supports inverted leds (which the NodeMCU is).

One important point here is that it is using a fork of WifiManager, because I needed a callback during the setup phase to flash the LEDs while waiting for setup.  This fork is hosted on my GitHub, and it is referenced from the platform.io (so you need to reload those dependencies).  I did a PR for the change to WifiManager, but it was rejected because they have a new feature coming that will work instead by making the Setup of WifiManager non-blocking.  I attempted to use that, but it depended on a development version of WifiManager, which in turn depended on new versions of ESP8266WebServer and all that together crashed and burned.  Plus, I didn't feel great about making it a requirement to be on a development branch.  When the new WifiManager is released, it is a simple change to adopt it instead of the callback in the fork.

I think this change will help in usability, especially during Wifi configuration and MQQT debugging.  I noticed there was an outstanding Issue asking for this - I will make a comment there as well.


chrismidgley:
Pulled too much.  Redoing it.

PR #227, opened 2018-02-16
Handle inconsistency on saturation/temp on v2 types

cmidgley:
Changes to saturation and temp changes on the v2 types (rgb+cct and fut089).  Detects if current bulb mode is unable to support the feature, and briefly switches modes, applies the change, and switches back.  This is especially important for temperature, otherwise it's hard to change temperature when using color mode. 

This may be a conceptual debate about the goals of MiLightHub... as this is changing the behavior of what the hub does vs. what a remote would do.  I think in the long run, it is a nice improvement as people use MQQT and the REST API to drive bulb changes, and this implements the functionality likely desired rather than pushing the limitations of the bulbs onto the user and their automation system.

### DIFF
--- a/lib/MiLight/FUT089PacketFormatter.cpp
+++ b/lib/MiLight/FUT089PacketFormatter.cpp
@@ -18,9 +18,25 @@ void FUT089PacketFormatter::updateBrightness(uint8_t brightness) {
   command(FUT089_BRIGHTNESS, brightness);
 }
 
+// change the hue (which may also change to color mode).  Also change saturation
+// if this was pending.
 void FUT089PacketFormatter::updateHue(uint16_t value) {
   uint8_t remapped = Units::rescale(value, 255, 360);
   updateColorRaw(remapped);
+
+  // look up our current mode 
+  GroupState ourState = this->stateStore->get(this->deviceId, this->groupId, REMOTE_TYPE_FUT089);
+
+  // do we have a saturation pending?
+  if (ourState.isPendingSaturation()) {
+    // now make the saturation change
+    command(FUT089_SATURATION, 100 - ourState.getSaturation());
+    ourState.setPendingSaturation(false);
+
+    // clear pending status
+    ourState.setPendingSaturation(false);
+    this->stateStore->set(this->deviceId, this->groupId, REMOTE_TYPE_FUT089, ourState);
+  }
 }
 
 void FUT089PacketFormatter::updateColorRaw(uint8_t value) {
@@ -33,8 +49,7 @@ void FUT089PacketFormatter::updateColorRaw(uint8_t value) {
 // back to the original mode.
 void FUT089PacketFormatter::updateTemperature(uint8_t value) {
   // look up our current mode 
-  BulbId ourBulb(this->deviceId, this->groupId, REMOTE_TYPE_FUT089);
-  GroupState ourState = this->stateStore->get(ourBulb);
+  GroupState ourState = this->stateStore->get(this->deviceId, this->groupId, REMOTE_TYPE_FUT089);
   BulbMode originalBulbMode = ourState.getBulbMode();
 
   // are we already in white?  If not, change to white
@@ -45,55 +60,31 @@ void FUT089PacketFormatter::updateTemperature(uint8_t value) {
   // now make the temperature change
   command(FUT089_KELVIN, 100 - value);
 
-  // revert back to the prior mode
-  switch (originalBulbMode) {
-    case BulbMode::BULB_MODE_COLOR:
-      updateHue(ourState.getHue());
-      break;
-    case BulbMode::BULB_MODE_NIGHT:
-      enableNightMode();
-      break;
-    case BulbMode::BULB_MODE_SCENE:
-      updateMode(ourState.getMode());
-      break;
-    case BulbMode::BULB_MODE_WHITE:
-      // no need to do anything, as we are wanting to stay in white
-      break;
+  // and return to our original mode
+  if (originalBulbMode != BulbMode::BULB_MODE_WHITE) {
+    switchMode(ourState, originalBulbMode);
   }
 }
 
 // change the saturation.  Note that temperature and saturation share the same command 
 // number (7), and they change which they do based on the mode of the lamp (white vs. color mode).
-// To make this command work, we need to switch to color mode, make the change, and then flip
-// back to the original mode.
+// Therefore, if we are not in color mode, we save the saturation and wait until we do a
+// mode change to color (hue) and then change the saturation.
 void FUT089PacketFormatter::updateSaturation(uint8_t value) {
   // look up our current mode 
-  BulbId ourBulb(this->deviceId, this->groupId, REMOTE_TYPE_FUT089);
-  GroupState ourState = this->stateStore->get(ourBulb);
+  GroupState ourState = this->stateStore->get(this->deviceId, this->groupId, REMOTE_TYPE_FUT089);
   BulbMode originalBulbMode = ourState.getBulbMode();
 
-  // are we already in white?  If not, change to white
-  if (originalBulbMode != BulbMode::BULB_MODE_COLOR)
-    updateHue(ourState.getHue());
-
+  // are we already in color?  If not, we can't make the change yet, so store the value
+  // so we can make it later
+  if (originalBulbMode != BulbMode::BULB_MODE_COLOR) {
+    ourState.setPendingSaturation(true);
+    ourState.setSaturation(value);
+    this->stateStore->set(this->deviceId, this->groupId, REMOTE_TYPE_FUT089, ourState);
+    return;
+  }
   // now make the saturation change
   command(FUT089_SATURATION, 100 - value);
-
-  // revert back to the prior mode
-  switch (originalBulbMode) {
-    case BulbMode::BULB_MODE_COLOR:
-      // no need to do anything, as we are wanting to stay in hue
-      break;
-    case BulbMode::BULB_MODE_NIGHT:
-      enableNightMode();
-      break;
-    case BulbMode::BULB_MODE_SCENE:
-      updateMode(ourState.getMode());
-      break;
-    case BulbMode::BULB_MODE_WHITE:
-      updateColorWhite();
-      break;
-  }
 }
 
 void FUT089PacketFormatter::updateColorWhite() {
--- a/lib/MiLight/FUT089PacketFormatter.cpp
+++ b/lib/MiLight/FUT089PacketFormatter.cpp
@@ -27,13 +27,73 @@ void FUT089PacketFormatter::updateColorRaw(uint8_t value) {
   command(FUT089_COLOR, FUT089_COLOR_OFFSET + value);
 }
 
+// change the temperature (kelvin).  Note that temperature and saturation share the same command 
+// number (7), and they change which they do based on the mode of the lamp (white vs. color mode).
+// To make this command work, we need to switch to white mode, make the change, and then flip
+// back to the original mode.
 void FUT089PacketFormatter::updateTemperature(uint8_t value) {
-  updateColorWhite();
+  // look up our current mode 
+  BulbId* ourBulb = new BulbId(this->deviceId, this->groupId, REMOTE_TYPE_FUT089);
+  GroupState ourState = this->stateStore->get(*ourBulb);
+  BulbMode originalBulbMode = ourState.getBulbMode();
+
+  // are we already in white?  If not, change to white
+  if (originalBulbMode != BulbMode::BULB_MODE_WHITE) {
+    updateColorWhite();
+  }
+
+  // now make the temperature change
   command(FUT089_KELVIN, 100 - value);
+
+  // revert back to the prior mode
+  switch (originalBulbMode) {
+    case BulbMode::BULB_MODE_COLOR:
+      updateHue(ourState.getHue());
+      break;
+    case BulbMode::BULB_MODE_NIGHT:
+      enableNightMode();
+      break;
+    case BulbMode::BULB_MODE_SCENE:
+      updateMode(ourState.getMode());
+      break;
+    case BulbMode::BULB_MODE_WHITE:
+      // no need to do anything, as we are wanting to stay in white
+      break;
+  }
 }
 
+// change the saturation.  Note that temperature and saturation share the same command 
+// number (7), and they change which they do based on the mode of the lamp (white vs. color mode).
+// To make this command work, we need to switch to color mode, make the change, and then flip
+// back to the original mode.
 void FUT089PacketFormatter::updateSaturation(uint8_t value) {
+  // look up our current mode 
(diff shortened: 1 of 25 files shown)


sidoh:
I agree this is bad behavior, and that changing into color mode is a good idea.

I sort of wonder if we should leave the device in color mode instead of switching back to white mode.  When white temperature is changed, the device changes to white mode and stays in white mode.  Thoughts?

cmidgley:
I think the question may be different for saturation than for temperature.  I'm not sure the saturation fix is as important, since it only seems to take effect in color mode (on my bulbs).  I can just see somebody making a small mistake with MQQT or REST where they set saturation then set color, and don't get the result they expected.  

But temperature I think makes sense to move back to the original mode.  On the V2 bulbs, it you are in color mode with some level of saturation that mixes in white, it will use the temperature.  But you can't change temperature without going to white first, changing temperature, and going back again.  This takes care of that.

lib/MiLight/FUT089PacketFormatter.cpp
```diff
@@ -27,13 +27,73 @@ void FUT089PacketFormatter::updateColorRaw(uint8_t value) {
...
 }
 
+// change the temperature (kelvin).  Note that temperature and saturation share the same command 
+// number (7), and they change which they do based on the mode of the lamp (white vs. color mode).
+// To make this command work, we need to switch to white mode, make the change, and then flip
+// back to the original mode.
 void FUT089PacketFormatter::updateTemperature(uint8_t value) {
-  updateColorWhite();
+  // look up our current mode 
+  BulbId* ourBulb = new BulbId(this->deviceId, this->groupId, REMOTE_TYPE_FUT089);
```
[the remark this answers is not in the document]
cmidgley: Great catch ... and it's a memory leak!  Will change...

sidoh:
Oh, I see what you're saying.

With the bulbs I have (FUT015 and FUT105), updating the white temperature already puts them into white mode.  Is that not the case with yours?

I actually think it makes a lot of sense to switch into color mode when the saturation is updated.  I think all I mean to say is that it makes slightly more sense for the bulb to stay in color mode rather than to switch back.

cmidgley:
No - temperature and saturation share the same ID (on FUT089), so their action is dependent upon the mode.  If color, it does saturation.  If white, it does temperature.  It's a bit different on RgbCct but my code goal was to make their behavior consistent independent of the type.

So basically if the bulb is white, it stays white after the change.  And if it is color, it stays color after the change.  But in both cases, the change is successfully sent to the bulb.  Therefore on RGBWW bulbs, it lets you change temperature while staying in your color (for example, blue with a mixed cool white or a mixed warm white).  When in white and changing saturation, it flips to color briefly, makes the change, and flips back.  This isn't so valuable, as the technically better approach would be to recall saturation and wait until you flip to color to execute the saturation change.  But this was easier!

lib/MiLight/FUT089PacketFormatter.cpp
```diff
@@ -27,13 +27,73 @@ void FUT089PacketFormatter::updateColorRaw(uint8_t value) {
...
+  // are we already in white?  If not, change to white
+  if (originalBulbMode != BulbMode::BULB_MODE_WHITE) {
+    updateColorWhite();
+  }
+
+  // now make the temperature change
   command(FUT089_KELVIN, 100 - value);
+
+  // revert back to the prior mode
+  switch (originalBulbMode) {
```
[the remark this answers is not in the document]
sidoh: I was actually just looking at the FUT089 formatter.  Seems like this code is here and in `updateSaturation`, with the caveat that the no-op case is different.  But seems like that'd be easy to abstract.

But you bring up a good point -- it actually seems like the base `PacketFormatter` should have an `switchBulbMode` method.  There are some leaky abstractions because not all device types support color, etc., but `PacketFormatter` already has `updateHue`.  What differences did you notice between formatters that made it seem like the abstraction would be difficult? 

I think you're right that it's not worth spending too much effort.

sidoh:
Aaah, right.  I don't know why I forgot that. :)

This is probably the source of my confusion -- the bulbs I have near my computer (and therefore the ones I test with most often) are FUT015s.  I don't think they have the behavior you're describing.  With lower saturation values, both the CW and WW LEDs are on with roughly equal brightness.

Here's an image with saturation 0, color temp 0 (coolest) --



This seems to suggest that the white temp state doesn't always cross the two modes, and that it might be cleanest to just keep them entirely separate (meaning just switch to and stay in the mode the command corresponds to before updating the state).

lib/MiLight/FUT089PacketFormatter.cpp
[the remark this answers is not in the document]
cmidgley: I can look into abstracting those temperature/saturation state changes if you want.  If so, it would want to be added to the base class V2PacketFormatter so it can be shared across V2 devices (I guess it could go all the way back to PacketFormatter, but I'm not sure if this state issue exists in non-V2 devices?).  But I'd rather not try to refactor RgbCct and FUT089 into some common base class at this time.

If we want to look at this - do you know if V1 devices have any of these state issues?  I don't think so, but thought you might know better than I!

cmidgley:
Not sure I follow your last paragraph.  Do you mean don't switch modes, and instead just flip and stay (for example, in in color and changing temperature, flip to white and stay there?).  That doesn't make sense to me, because that isn't what the bulb does today - it doesn't flip to white, but instead adjusts saturation.  You had code in there to flip it to white, which made it unfriendly when trying to change temperature while in color.  Either way the code was making a change - I'm proposing the change is, IMHO, more user friendly to expected action.

What is the negative side effect on your bulb with this change?

sidoh:
Let me back up and detail what I have in mind.  If it still feels like there's a disconnect, I've probably misunderstood something, and I'll go back and look at things more carefully.

1. A command sent to REST/MQTT of the form `{"saturation":50}` should:
   * Switch the bulb into color mode, if it's not already in color mode
   * Set saturation to 50
   * Nothing else -- should stay in color mode
2. A command sent to REST/MQTT of the form `{"kelvin":0}` should:
   * Switch the bulb into white mode, if it's not already in white mode
   * Set kelvin to 0
   * Nothing else -- should stay in white mode

I say this because:

* (2) is already the behavior for FUT015 bulbs.  In fact the only way to enter white mode is to send a set kelvin packet
* white temp only has an effect in white mode for FUT015 bulbs.  White temperature isn't adjustable in color mode.

cmidgley:
I think the perspective difference is:

1) We want to emulate the behavior of the bulbs, while adding missing functionality as appropriate
2) We want to fix the (poor) behavior of the bulbs, even though it is different than how a controller might behave

I've been leaning towards #2 - make the bulbs easier to use.

My specific issue that drove this whole deal was simply this:

1) I'm using an RGBWW bulb, and I am in a color (say, Blue).
2) I reduce saturation, causing the white bulbs to engage
3) I don't like the white color, so I change the temperature.

This doesn't work with the bulb naturally.  You can't change temperature without moving to white, losing your color settings.  It's annoying to use these bulbs and shifting your white temperature around as a result.

If we take your proposal, I end up with white and now I have to reset my color again (extra painful when using a wifi app/web site, as now I need to try to find that nice color again).  My proposal changes the temperature, even though it comes at the cost of a brief flash of white light while the bulb transitions into white, makes the change, and transitions back.

Also, there are more modes than color and white.  There are the "disco" modes (what do you call those?) and the night mode.  Currently I always revert to those modes after applying the change (though I think I just saw a bug on the night mode).  This makes the setting "consistent" so that any change to saturation is just a change to saturation - it never changes the bulb mode.  Same with temperature - while it changes temperature, it never changes mode.  It just seems odd to me, even though some bulbs do this, to change bulb mode when changing temperature or saturation.

Having said all that, we can instead say it's ok for saturation to change bulb mode.  I don't think this is a very important issue as it doesn't really do much.  But I do feel that it is odd behavior of the bulbs (poor user experience) to change mode simply by changing requested temperature.

If you don't agree, no worries.  We can easily do whatever you feel is best, and I'll support that change.

sidoh:
I fervently agree with this philosophy:

> We want to fix the (poor) behavior of the bulbs, even though it is different than how a controller might behave

I have no interest in emulating silly behavior unnecessarily. :)

I think perhaps the disconnect is that I didn't know that white temperature was adjustable in color mode on some devices.  With the bulbs I test with primarily (FUT015), it is not possible to do this -- CW and WW have a fixed brightness.  

This makes me feel like the mode-switching behavior is not intuitive for FUT015s (these are the 8W earlier generation RGB+CCT bulbs).  But it sounds like it is the most intuitive for FUT105s (the 12W newer generation bulbs that support the FUT089 protocol).

So maybe the right approach is to do what you're suggesting in the FUT089 protocol, but not the RGB+CCT (FUT092) protocol.

Maybe the most intuitive approach is to never switch modes on state update commands.  For example, if we received the command `{"hue":0}`:

1. If already in color mode, make the adjustment (e.g., set hue to 0)
2. If in a different mode, update the internal state (set hue to 0), but don't send the RF packet
3. When switching to a different mode, send packets to put the bulbs in a state consistent with internal state

In the case of the FUT089 protocol, we would support updating white temperature while in color mode using the workaround you're suggesting.

cmidgley:
That's a reasonable solution, though will take some more thinking.  Let me noodle on that for a bit.

I think it is a bit strange, but not totally, that we would offer different behaviors on the V2 devices depending on which type is used (FUT089 vs. RgbCct).  But I guess that's not all that inconsistent with the stupid protocol changes MiLight makes with each generation of their product!

cmidgley:
Getting closer here.  Changed to supporting saturation change only upon hue changes.  Also did some refactoring to clean up redundant / unclear code.  

The one thing I have not yet addressed is if FUT089 should be different than RGB_CCT in it's behavior.  For the bulb I use, they behave the same - it's just a four group vs. eight group type.  If there are other bulbs that behave differently on those two types, then maybe it does make sense to change?  I'm still unclear on this, as I don't experience it on my bulb.  Therefore, at the moment (easy to change) RGB_CCT continues to support changing temperature while in hue mode.

cmidgley:
On further reflection, I think it is fine to have RGB_CCT behave differently.  Not so much because I think the type (of a remote) should define behavior, but more because it is a way (maybe not the best) to differentiate functionality.

Having said that ... another answer here is we add this as a settings property.  We could have a property called something like "automatically adjust temperature and saturation" that when set to a 1 would add this functionality, and otherwise it would not.  Then we can choose whatever default we want (0 for existing behavior, 1 for new behavior).   To do this, we would either need to pass Settings into the constructor (in PacketFormatter) like I added for GroupStateStore, or we would want to make Settings into a singleton so it can be referenced anywhere.  My feeling is that it belongs as a singleton since it is a shared concept for behavior, but some people don't like the [singleton pattern](https://en.wikipedia.org/wiki/Singleton_pattern) as it is a form of a global.

One more thing - I just noticed when looking at what it would take to implement this idea that we pass GroupStateStore into parsePacket, yet now that it is provided by the constructor I should remove that parameter as it is redundant an unnecessary.  Agree?

cmidgley:
Here is a "no harm done" update.  This update uses Settings to determine if this automated mode switching feature is used or not, and by default, it is off.  Since Settings is now passed to PacketFormatter (and subclasses), it also let me refactor some of the code where some settings (or the entire settings object) were passed in, and are no longer necessary.

I have not done sufficient testing, and will try to test more deeply later this weekend or early next week.  But I thought I'd put it up and see what thoughts you had on this as an approach.

cmidgley:
Minor update as I did testing.  When using the FUT104 bulb (MR16, RGB+CCT bulb) and using the RGB+CCT type, changing temperature switches to white.  But when using the FUT089 type, changing temperature just changes saturation.  This is because the protocol shares saturation and temp on the FUT089 adn the FUT014 can accept both protocols.  I think before you handled this by changing to white when changing temperature on the FUT089 - so I put that back in (when not using the new automated mode switching feature).

Also changed to using a cleaner radio button in settings for enable/disable rather than 1/0 or true/false being typed in.

lib/MiLight/V2PacketFormatter.cpp
```diff
@@ -80,3 +80,23 @@ void V2PacketFormatter::format(uint8_t const* packet, char* buffer) {
 uint8_t V2PacketFormatter::groupCommandArg(MiLightStatus status, uint8_t groupId) {
   return GROUP_COMMAND_ARG(status, groupId, numGroups);
 }
+
+// helper method to return a bulb to the prior state
+void V2PacketFormatter::switchMode(GroupState currentState, BulbMode desiredMode) {
```
sidoh: Perhaps we should skip doing an update if `currentState.getBulbMode() == desiredMode`?

lib/MiLight/V2PacketFormatter.cpp
```diff
@@ -80,3 +80,23 @@ void V2PacketFormatter::format(uint8_t const* packet, char* buffer) {
...
+      break;
+    case BulbMode::BULB_MODE_NIGHT:
+      enableNightMode();
+      break;
+    case BulbMode::BULB_MODE_SCENE:
+      updateMode(currentState.getMode());
+      break;
+    case BulbMode::BULB_MODE_WHITE:
+      updateColorWhite();
+      break;
```
sidoh: Might be good to have a `default` case that prints a warning.

lib/MiLightState/GroupState.cpp
```diff
@@ -297,6 +298,17 @@ bool GroupState::setNightMode(bool nightMode) {
   return true;
 }
 
+bool GroupState::isPendingSaturation() const { return state.fields._isPendingSaturation; }
```
sidoh: Now that it's optional, I think it's fine for it to behave the way you had it in the original PR.

The concerns I have with attaching it to state are:

* We're having to add another byte, which might cause issues with backwards compatibility
* This isn't really a persistent state field, so it feels a little weird putting it here
* Because the field is ignored when loaded, things might behave strangely if a bulb's state is flushed to flash (a limited number of states are kept in RAM)

It feels like if we want to go down this road, we should do it for every field, and I think the approach we'd take would look a little different if we did.

lib/MiLight/V2PacketFormatter.cpp
cmidgley: This doesn't work because the current state isn't changed right away, causing the comparison to be invalid.  I believe it doesn't change until the packet round-trips, which is after the inline code executes.

cmidgley:
All recommended changes, minus the one that doesn't work, have been made.  Let me know if anything is broken, or if other changes are needed.

lib/MiLightState/GroupState.h
```diff
@@ -115,6 +115,7 @@ class GroupState {
         _isSetState           : 1,
         _isSetHue             : 1;
       uint32_t
+        _isNightMode          : 1,
```
sidoh: Why are we moving this?  I think any changes to the ordering of the fields here will break backwards compatibility.

cmidgley:
I saw a comment in email asking why the StateData union had changed order, but I don't see it here?  Anyway - that was a botched backed of the union changes on my my part.

I have reverted to the original union and pushed it here.

PR #228, opened 2018-02-16
Improve state debugging

cmidgley:
Currently device type is compared for state, even though only device ID and group identify a bulb.  On some bulbs (the more modern ones, with RGBWW) support multiple controllers and switching between them is confusing when state is different across device types.  For me, this is a usability issue - it is easier for me to use when state remains the same when switching between remote types.  

The larger change here is some nice debugging code to help debug state changes.

### DIFF
--- a/lib/MiLightState/GroupState.cpp
+++ b/lib/MiLightState/GroupState.cpp
@@ -52,10 +52,8 @@ void BulbId::operator=(const BulbId &other) {
 // (type of controller/remote) as this doesn't directly affect the identity of the bulb
 bool BulbId::operator==(const BulbId &other) {
   return deviceId == other.deviceId
-    && groupId == other.groupId;
-    // used to contain this test below for deviceType, removed as it does not identify the bulb, but instead
-    // identifies which remote is being used
-//    && deviceType == other.deviceType;
+    && groupId == other.groupId
+    && deviceType == other.deviceType;
 }
 
 GroupState::GroupState() {
--- a/lib/MiLightState/GroupState.cpp
+++ b/lib/MiLightState/GroupState.cpp
@@ -490,10 +490,9 @@ void GroupState::applyField(JsonObject& partialState, GroupStateField field) {
 
 // helper function to debug the current state (in JSON) to the serial port
 void GroupState::debugState(char const *debugMessage) {
-#ifdef DEBUG_PRINTF
+#ifdef DEBUG_STATE
   // using static to keep large buffers off the call stack
-  static char buffer[1000];
-  static StaticJsonBuffer<1000> jsonBuffer;
+  static StaticJsonBuffer<500> jsonBuffer;
 
   // define fields to show (if count changes, make sure to update count to applyState below)
   GroupStateField fields[] { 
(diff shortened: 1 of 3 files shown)


sidoh:
The thinking here is that some bulbs can use multiple device types, and keying on device type when storing state means there's more than than one source of truth?

On the flip side:

* not all devices do this
* I don't think any device supports all device types.  

The FUT105s I have, for example, support both FUT089 and FUT092 protocols, but:

* it "binds" to whichever it's paired with.  
* for example, if you pair with a FUT089 remote, the bulb doesn't respond to FUT092 commands from the same device ID.  
* you _can_ pair with both a FUT089 and a FUT092 having the same device ID, but you have to do so separately.

This kind of behavior is why device type is incorporated in the state key.  Seems like both behaviors exist, but I'm not sure how to resolve that.

sidoh:
I really like the debugging changes and comments you added, so we should get those in regardless of what we decide here.

I agree that the best solution from a user's perspective would allow for customizable behavior here.  But even then, there probably has to be a default behavior, and it should be whatever makes most sense. :)

If we end up going this route, we might want to consider making some changes to state persistence, as it's keyed on device type as well.

sidoh:
I'd like to get some thoughts from other users.  Pulling arbitrarily from the list people who have contributed in the past:

@DennisVonDerBey, @garmck, @fliphess, @tbrasser -- any opinions here?  The basic question is whether devices with the same (device id, group id) pair should share the same state, regardless of device type.

cmidgley:
I was not aware, prior to our discussions, that some bulbs bind differently to different device types.  Do you know if that is specific to device types ... meaning, do the v2 devices not do this (mine don't) yet the v1 devices may?  If so, maybe we determine if type is part of compare based on if we are switching between v2 devices?

.  I also did not notice in state where this type was used as a selector of state (though didn't spend a lot of time looking) - I just saw how it used BulbId to make that basic decision.

Honestly - in the end - this isn't very important.  We can easily back out the bulbid == change and keep the debugging.

cmidgley:
I put back the device type compare, so this is now just the debugging code.  That change doesn't matter a lot to me, so I'd rather just get this one wrapped up.

cmidgley:
Sorry - didn't mean to close it.  Reopen!

lib/MiLightState/GroupState.cpp
```diff
@@ -462,6 +488,43 @@ void GroupState::applyField(JsonObject& partialState, GroupStateField field) {
   }
 }
 
+// helper function to debug the current state (in JSON) to the serial port
+void GroupState::debugState(char const *debugMessage) {
+#ifdef DEBUG_PRINTF
+  // using static to keep large buffers off the call stack
```
sidoh: I think we might want to not use `static` here.  I think that this will allocate 2KBs of RAM globally, instead of just when this method is called.  We really only have 10-20 KBs of RAM to play with before things start getting weird.

cmidgley:
It can't live on the stack - it blows chunks if you do!  We could dynamically allocate and free each time, but I thought static was fine since it is only allocated if you have enabled debugging.  Let me know...

sidoh:
Ah, I missed that.  Think you're right, no reason to mess with it if putting it on the stack doesn't work.

You might be able to do `jsonBuffer.printTo(Serial)`.  Maybe that let us kill the second buffer.  But not a big deal since this is only there when debugging is enabled, as you point out. :)

cmidgley:
I'll give the jsonBuffer.printTo approach a try...

cmidgley:
Here is an update that reduces allocated size from 2K to 500 bytes (I'm seeing the largest states somewhere around 350 bytes).  I also renamed DEBUG_PRINTF to DEBUG_STATE so we can be more granular when using it.

sidoh:
Awesome! 👍

PR #229, opened 2018-02-16
Add LED status to indicate wifi condition and packet handling

cmidgley:
This PR includes a new feature to update an LED with status of the device. See the README.md for details on the different flashing modes. It indicates if the Wifi is in AP mode, connected or failed, as well as signals when packets are traveling. The UI has been modified to support setting the pin of the LED, but it defaults to pin 2 (the built-in blue LED on NodeMCU) and supports inverted leds (which the NodeMCU is).

One important point here is that it is using a fork of WifiManager, because I needed a callback during the setup phase to flash the LEDs while waiting for setup. This fork is hosted on my GitHub, and it is referenced from the platform.io (so you need to reload those dependencies). I did a PR for the change to WifiManager, but it was rejected because they have a new feature coming that will work instead by making the Setup of WifiManager non-blocking. I attempted to use that, but it depended on a development version of WifiManager, which in turn depended on new versions of ESP8266WebServer and all that together crashed and burned. Plus, I didn't feel great about making it a requirement to be on a development branch. When the new WifiManager is released, it is a simple change to adopt it instead of the callback in the fork.

I think this change will help in usability, especially during Wifi configuration and MQQT debugging. I noticed there was an outstanding Issue asking for this - I will make a comment there as well.

### DIFF
--- a/lib/LEDStatus/LEDStatus.cpp
+++ b/lib/LEDStatus/LEDStatus.cpp
@@ -1,0 +1,172 @@
+#include "LEDStatus.h"
+
+// constructor defines which pin the LED is attached to
+LEDStatus::LEDStatus(int8_t ledPin) {
+    // if pin negative, reverse and set inverse on pin outputs
+    if (ledPin < 0) {
+        ledPin = -ledPin;
+        _inverse = true;
+    } else
+        _inverse = false;
+    // set up the pin
+    _ledPin = ledPin;
+    pinMode(_ledPin, OUTPUT);
+    digitalWrite(_ledPin, _pinState(LOW));
+    _timer = millis();
+}
+
+// change pin at runtime
+void LEDStatus::changePin(int8_t ledPin) {
+    bool inverse;
+    // if pin negative, reverse and set inverse on pin outputs
+    if (ledPin < 0) {
+        ledPin = -ledPin;
+        inverse = true;
+    } else
+        inverse = false;
+
+    if ((ledPin != _ledPin) && (inverse != _inverse)) {
+        // make sure old pin is off
+        digitalWrite(_ledPin, _pinState(LOW));
+        _ledPin = ledPin;
+        _inverse = inverse;
+        // and make sure new pin is also off
+        pinMode(_ledPin, OUTPUT);
+        digitalWrite(_ledPin, _pinState(LOW));
+    }
+}
+
+
+// identify how to flash the LED by mode, continuously until changed
+void LEDStatus::continuous(LEDStatus::LEDMode mode) {
+    uint16_t ledOffMs, ledOnMs;
+    _modeToTime(mode, ledOffMs, ledOnMs);
+    continuous(ledOffMs, ledOnMs);
+}
+
+// identify how to flash the LED by on/off times (in ms), continuously until changed
+void LEDStatus::continuous(uint16_t ledOffMs, uint16_t ledOnMs) {
+    _continuousOffMs = ledOffMs;
+    _continuousOnMs = ledOnMs;
+    _continuousCurrentlyOn = false;
+    // reset LED to off
+    if (_ledPin > 0)
+        digitalWrite(_ledPin, _pinState(LOW));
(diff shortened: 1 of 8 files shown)


lib/LEDStatus/LEDStatus.h
```diff
@@ -0,0 +1,44 @@
+#include <Arduino.h>
+
+#ifndef _LED_STATUS_H
+#define _LED_STATUS_H
+
+class LEDStatus {
+    public:
+        enum LEDMode {
```
sidoh: Should we make this an `enum class`?  My guess is that your C++ is stronger than mine, so please feel free to educate me if there's not a good reason for this.

lib/LEDStatus/LEDStatus.cpp
```diff
@@ -0,0 +1,172 @@
...
+        case LEDMode::On:
+            ledOffMs = 0;
+            ledOnMs = 1000;
+            break;
+        case LEDMode::Flicker:
+            ledOffMs = 50;
+            ledOnMs = 30;
+            break;
+        default:
+            Serial.printf("LEDStatus::_modeToTime: Uknown LED mode %d\n", mode);
```
sidoh: should this be `Unknown`?

sidoh: Think we should probably put the string in PROGMEM.

src/main.cpp
```diff
@@ -237,20 +250,43 @@ bool shouldRestart() {
...
+  // tell Wifi manager to call us during the setup.  Note that this "setSetupLoopCallback" is an addition
+  // made to Wifi manager in a private fork.  As of this writing, WifiManager has a new feature coming that
+  // allows the "autoConnect" method to be non-blocking which can implement this same functionality.  However,
+  // that change is only on the development branch so we are going to continue to use this fork until
+  // that is merged and ready.
+  wifiManager.setSetupLoopCallback(handleLED);
+  wifiManager.setConfigPortalTimeout(180);
+  if (wifiManager.autoConnect(ssid.c_str(), "milightHub")) {
+    ledStatus->continuous(LEDStatus::LEDMode::SlowBlip);
+    Serial.printf("Wifi connected succesfully\n");
```
sidoh: Think we should probably put the string in PROGMEM.

src/main.cpp
```diff
@@ -237,20 +250,43 @@ bool shouldRestart() {
...
+  } else {
+    ledStatus->continuous(LEDStatus::LEDMode::On);
+    Serial.printf("Wifi failed.  Oh well.\n");
```
sidoh: Think we should probably put the string in PROGMEM.

sidoh:
I recognize this is nitpicky, but I would like to keep formatting consistent throughout the project. Could we --

1. Use 2 space indentation
2. Use 1TBS braces (just adding braces around one-line `if`s.)

lib/LEDStatus/LEDStatus.h
cmidgley: Yes, that is better.  I'm an old school C++ guy (haven't used C++ in countless years) so getting back to speed with this project!  I'll make that change.

lib/LEDStatus/LEDStatus.cpp
cmidgley: Ah!  That's what F(xxx) does!  Will change.

sidoh: In case this saves you any time -- the syntax is slightly different with printf.  You'd use `printf_P` and the `PSTR` macro.

cmidgley:
Here is an update that has the F() calls, the enum class, and supported your anal tendencies by changing to 2 spaces tabs and braces...!

sidoh:
Looks awesome! 👍

sidoh:
@cmidgley I just botched this one.  Had meant to change the base to v1.7.0, but forgot, and merged into master.

Can you re-open a PR for this one?  Really sorry about that :(

PR #231, opened 2018-02-16
Switched to using library ESP8266WebServer

chrismidgley:
Refactored code to support library ESP8266WebServer (instead of the local copy).  Reworked authentication to support the library without the hacked functions.  I did my best to test all functionality (device control, firmware upload, settings save/restore, sniffing, changing authentication, etc) but I don't have MQQT so I did not testing with that (I don't think I changed anything that should affect it, but testing is always a Good Thing).


sidoh:
Ah right, I'd forgotten about that.  I handled it in exactly the same way in [another project](https://github.com/sidoh/epaper_templates/blob/master/lib/HTTP/EpaperWebServer.cpp#L408-L429), and felt really good about the approach.  Meant to create a ticket over here to do that for espMH as well, but didn't get around to it.

This is much nicer!  Going to have a closer look at the code soon (hopefully this weekend) and we can get this merged.  

Thank you very much for taking the time to do this.

lib/WebServer/MiLightHttpServer.cpp
```diff
@@ -11,74 +11,31 @@
...
-      Settings::load(settings);
-      server.send_P(200, TEXT_PLAIN, PSTR("success."));
-    },
-    handleUpdateFile(SETTINGS_FILE)
-  );
-  server.on("/radio_configs", HTTP_GET, [this]() { handleGetRadioConfigs(); });
+  // set up HTTP end points to serve
+
+  _handleRootPage = handleServe_P(index_html_gz, index_html_gz_len);
+  server.on("/", HTTP_GET, [this]() { if (server.validateAuthentiation()) _handleRootPage(); });
```
sidoh: Can we pull the auth handling wrapper into a method so that the conditional isn't repeated?  I think probably the easiest way is to just add private methods for `on` and `onPattern` and use those in place of `server.on`, etc.

Example here, in case that's helpful:

https://github.com/sidoh/epaper_templates/blob/master/lib/HTTP/EpaperWebServer.cpp#L408-L429

This is using a different HTTP server, but should be the same idea.

cmidgley:
Refactored for the shared authentication logic.  Sorry about the styling, I thought I caught them all.  It's not a natural format for me.  If you find any other places, just point them out (as I have to scan the code by hand to find them right now)

sidoh:
This is awesome :)

PR #246, opened 2018-03-12
blip when connection failed, solid when successful

sidoh:
@cmidgley what do you think about this?

It feels like a solid on LED is a better success indicator than a blinking one.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -256,7 +256,7 @@ void handleLED() {
 void setup() {
   Serial.begin(9600);
   String ssid = "ESP" + String(ESP.getChipId());
-  
+
   // load up our persistent settings from the file system
   SPIFFS.begin();
   Settings::load(settings);
@@ -278,10 +278,10 @@ void setup() {
   wifiManager.setSetupLoopCallback(handleLED);
   wifiManager.setConfigPortalTimeout(180);
   if (wifiManager.autoConnect(ssid.c_str(), "milightHub")) {
-    ledStatus->continuous(LEDStatus::LEDMode::SlowBlip);
+    ledStatus->continuous(LEDStatus::LEDMode::On);
     Serial.println(F("Wifi connected succesfully\n"));
   } else {
-    ledStatus->continuous(LEDStatus::LEDMode::On);
+    ledStatus->continuous(LEDStatus::LEDMode::SlowBlip);
     Serial.println(F("Wifi failed.  Oh well.\n"));
   }
 


bombcheck:
I agree. "Blinking" gets more attention so it is better to show an error. To bad the led is not rgb. :)

cmidgley:
I can change it however we want, but I find bright doesn't tell me if it is working or not.  When the unit crashes, it can be solid on (or solid off) yet a slight blip shows that it is alive.  At least that was my thinking after I tried several combinations (and had some crashes in the middle!)

sidoh:
Ah, that makes sense.  Perhaps we should make it an option?

sidoh:
Was thinking of having it be a matrix.  Each state has a configurable LED behavior.  Maybe that's overdoing it, though.

Totally agree.  The entire UI desperately needs an overhaul.  Frankly I had meant for it to be a prototype, but it ended up not getting rebuilt.  Splitting settings into sections and not having them be inline would be a good incremental improvement.

cmidgley:
The current setup UI isn't great for doing a matrix.  As you know, it's more about asking single questions.  So we could do a simple ON/OFF switch for "Always on LED when operating" (and off would be blinking).  Or we could do a dropdown for each state, such as "LED when in AP WIFI mode", "LED when operating", and "LED when receiving/transmitting" and then choose from the set of standard LED modes that I've defined (or adjust those accordingly). 

If we want to get really fancy, then we could look at doing different pins based on states as well, and even support an external RGB LED for color states.  Not sure how many people would both to wire that up though.

I do have to wonder if this is an area worthy of further investment (at least until we get feedback from others)?

sidoh:
Yeah, there'd be a fair amount of UI code to support the matrix.  But if that's the best way to present choices to the user, I don't mind adding it.

re: RGB LED, I like the idea.  We'd need another 3 GPIO pins, right?  Feels like it'd be a bit of a heavyweight feature that might not get much use.

I don't have strong opinions on this.  I'd personally prefer for solid to be the normal state, but if I'm in the minority I don't mind leaving it as-is.

One thing that may be worth adding: hopefully crashes aren't something end-users have to deal with very frequently. :)

cmidgley:
One more thing ... right now, when the device is in AP mode but times out (never gets a client mode connect to Wifi) it currently turns solid ON to indicate a total failure.  So we would need to change that as well to some other condition if we wanted to have ON to mean all is working.  In that case, would it just be OFF?

I've started working on a small UI change to support a solid LED mode.  Do you want that to be a pull request from branch v1.7.0 (my preference since that branch is more complete), or back on the original LED branch (which I can easily do).

sidoh:
To me, blinking at regular intervals indicates an abnormal state: either setup (i.e., a router connecting) or failure.

If I had to make this decision in a vacuum, I think I'd go with:

1. Everything working normally: Solid on
1. Connection pending, setup mode: Slow blink (0.5-1x/second)
1. Connection failed or disconnected: Fast blink (2x/second)
1. Packet sent or received: Very fast blink (5x/second or more)

PR against the v1.7 branch makes most sense to me.  Whatever's easiest :)

cmidgley:
My view is...

- Slow consistent flashing means things are likely ok, like a battery charging or things moving alright.
- Fast consistent flashing means something is wrong - needs my attention
- Occasionally blips of light (not just on/off/on/off) means it's alive and working - a heart beat
- Solid on and solid off are ambiguous.  Can be ok, can be error? 

I also find multiple speeds of flashing is hard to recall / discern.  Very slow, slow and fast to indicate different conditions requires timing and counting.

In my configuration, the one I don't like is solid on indicating the wifi failure.  But I like it better than a different blink speed, because it's easy to differentiate and check against the readme.

Anyway - I think we have two choices here:

1) Just agree on one way or the other, it's just not that big of a deal.
2) Add a UI button to change behavior.  This is implemented (though I don't quite have the codes you recommend above, but can easily add them).  This does make the readme a bit more complex, as it has to describe two different flash codes based on this option.

What do you want?  Either way is the same amount of effort at this point.

sidoh:
Makes sense to me.  A very quick blip every 5-10s or so seems reasonable.

I think solid on or off only seem ambiguous to me in the case of a crash (maybe i'm missing something, though).  My feeling is that we should not optimize for this, because we should strive for crashes to never happen.  If different behavior makes development easier because there's a visual heartbeat, an option sounds great.

I think it'd be nice to have it be configurable. I agree that it makes things a little more complicated, but seems worthwhile since this is really a matter of preference. :)

Really i don't have strong opinions here.  let's just do whatever you think is best.

hastarin:
I certainly agree it's a matter of preference.  I already have enough devices I've had to employ electrical tape to block obnoxious LEDs!

Personally I certainly would prefer no LED at all when everything is working.  I've got a battery for backup and any LED would just waste power.  

Having some pattern during setup/error states makes perfect sense to me.

sidoh:
Think everything else for the 1.7 release is wrapped up other than this.

@cmidgley, sounded like you have some changes that expose settings for the LED behavior.  Do you feel like those are ready?  I'm also happy to take a stab.

cmidgley:
Sorry for my long absence.  I do have some changes but a few bugs that I need to work out.  Hopefully I'll be back soon with some updates.

cmidgley:
I think we can close out this pull request - see the other one I just created.

sidoh:
👍 sounds good.  Other PR looks awesome!

PR #252, opened 2018-03-19
Update index.html

olskar:


olskar:
Hm, i am very new to this. Tried to make the mode function more descriptive, dont know if its the right place

bombcheck:
Nice, but afaik the modes depend on the bulb type. For example I have some MiLight "wine glass" lamps with build-in battery. These have a "candle flicker" mode as the first mode.

olskar:
Ah, ok :)

Issue #23, opened 2017-03-29
Rebuild web UI with webpack, react, etc.


sidoh:
The current webapp is functional, but difficult to maintain.

PR #259, opened 2018-04-15
LED and UI changes

cmidgley:
Fairly big update - hope you like it...! 

I added support for LED status changes to let you have total control over the status of the LED in all the various conditions.  This required adding five new fields to setup.  When I did it, I thought the UI was getting too busy, so I went ahead and did a change to the UI to support settings in a modal along with tabs, fixed settings to use friendly names (instead of the internal variable names), cleaned up gateway servers to modal as well, cleaned up sniffing to use a nice bootstrap inline display (that can be dismissed), and did some nice styling on the UI to make it a cleaner experience.

Sorry this took so long...!


sidoh:
This looks really great.  I'll hopefully have a chance to test this by next weekend at the latest.

Thanks for improving the UI.  It was definitely getting super messy.

rjsachse:
Oh good work i will have to have a look as i have been changing the ui too, seems like you have done more than i planed

sidoh:
Perhaps worth mentioning that I've been doing a little embedded UI modernization for a different project.  Major differences are:

1. Use webpack instead of gulp.  Pulling in libraries from npm rather than random CDNs
1. Produces a minified HTML blob with all javascript inlined (including libraries)
1. More modern web frameworks (React/redux).  Less manual state transformations and such
1. Generally far less spaghetti-ish.  ES2015 and separate modules go a long way

Still a work in progress at the moment, but you can see it here:

https://github.com/sidoh/epaper_templates/tree/master/web

I'd like to pull it into this project at some point.  Actually meant to do it far earlier (see #23 lol), but haven't gotten around to looking into webpack until recently.

In any case, definitely still want to get this PR merged.  Still planning on looking at it more carefully this weekend.

web/src/js/script.js
```diff
@@ -4,16 +4,219 @@ var UNIT_PARAMS = {
...
+    tag: "tab-setup",
+    friendly: "Setup",
+  }, {
+    tag: "tab-led",
+    friendly: "LED",
+  }, {
+    tag: "tab-radio",
+    friendly: "Radio",
+  }, {
+    tag: "tab-mqqt",
```
sidoh: Should be MQTT, not MQQT :)

Think MQQT shows up in a couple of places.

web/src/index.html
```diff
@@ -137,14 +137,11 @@ <h2 class="modal-title">Restore Settings</h2>
     </div>
   </div>
 
+  <nav class="navbar navbar-expand-lg">
+    <a class="navbar-brand" href="https://github.com/sidoh/esp8266_milight_hub">MiLight Hub</a>
```
sidoh: I am used to the brand link pointing at `/` (e.g., being equivalent to a refresh).

Linking to the Github page seems like a good idea.  Maybe a pulled right Github logo would make more sense?

web/src/index.html
```diff
@@ -367,107 +364,120 @@ <h5>Commands</h5>
...
+      <button type="button" class="close" id="traffic-sniff-close">&times;</button>
+      <h4>Traffic sniffed (sent and received)</h4>
+      <div id="sniffed-traffic"></div>
     </div>
+  </div>
 
-    <div class="row header-row">
-      <div class="col-sm-12">
-        <h1>Admin</h1>
+  <footer>
```
sidoh: I think my preference would be to have at least some of these in the nav bar, styled as links rather than buttons.

Perhaps some of the admin type actions should be buttons in a separate modal.

cmidgley:
Well, that proves I don't use MQTT!  Easy...

I agree that a web page, on the net, often uses root for that upper-left graphic.  I thought that didn't make much sense or add much value, and I thought linking to your github page (or maybe your blog?) would make more sense.   Truthfully, none of it makes much difference so could just remove the link if it bothers you.

The nav bar change I also wondered about - improving overall navigation.  But that was a slippery slope - what goes to the top vs. lower?  Or hidden in some modal (I thought about putting some on the setup dialog).  In the end, I decided to not try to tackle that and engage in the debate that would surely follow (but... I guess we are there anyway!)  If you have a specific suggestion, glad to look at it (or obviously feel free to do it yourself).  Could even go as far as a menu system with sub-links to further simplify... 

I'll make the MQTT change shortly and await your thoughts on anything else.

rjsachse:
I tried the 1.7.0-dev7 firmware and getting some strange results.

If i log on to the hib via my old s2 phone it works but if i try on chromium or chrome on my note 4 i just get a blank page, on chromium the loading symbol just keeps spinning.

I have so far formatted my nodemcuv2 so many times, tried to compile from code and the release bin, nothing seems to have helped. I tried all version of 1.7.0-dev1 to 7 and all are fine except v7.

Oh and have cleared caches on browsers to

southwalesboy:
Same issue here. Chrome on Android kicks up a blank page but firefox on Android works fine and looks great!

sidoh:
Thanks for the heads up guys.  Will take a look.  dev7 was a big UI change.  Would recommend reverting to dev6 for now.

Issue #265, opened 2018-04-27
esp8266 bugged state


jstans:
Wish I had a better name but I'm really not sure what's happening here.

So, I've just set up my ESP8266 with compiled source from the master branch and it is now working with my CCT bulbs (and it's great!) but I'm having a strange issue that I'm now trying to diagnose. I am using Home Assistant with MQTT and the esp8266 seems to get in this state where it seems to stop sending any commands and seems to get confused with the state it sends out over MQTT. This is fixed temporarily with a restart of the esp8266 until it happens again.

A reliable way of entering the state seems to be to try and set a warm colour on home assistant though I am confident there are other ways. Currently assuming this is sending an out of range value or something which triggers the bug. After this no commands work and weirdly when using a remote to send commands it sniffs the packet correctly but sends the MQTT update for the wrong group.

I should also note that despite the group_state_fields the only fields I see in the MQTT messages are state, bulb_mode and color.

sidoh:
I think this is probably because CCT brightness/temp commands involve 10x or more packets than are usually required.  This is basically because it has to be decomposed into a series of up/down commands.

From my experience, when espMH tries to send too many packets at once, the TCP stack gets overwhelmed and it gets into a funky state (sounds like this is what you're describing).

I'd recommend running through the suggestions here:

https://github.com/sidoh/esp8266_milight_hub/wiki/Performance-Tuning

In particular for your situation:

* Make sure `state_flush_interval` is set pretty high.  Probably something like 1s.
* Disable (clear) `mqtt_update_topic_pattern`.
* Enable packet repeat throttling.  This feature is sort of difficult to navigate, but will probably help a lot.  Make sure `packet_repeat_minimum` is set to something high enough that your bulbs are getting to the right state.

sidoh:
Yes, it's done this way due to the way the CCT protocol is implemented.  There isn't a "set brightness to X" command, so this is the only way to achieve setting brightness directly.

Ah, I neglected to ask if you were on 1.7.  #170 fixed a bug with exactly this functionality.

Keep in mind that repeat throttling is disabled by default.  Set throttle sensitivity to something >0 to enable it.

MQTT state only includes fields with known values.

jstans:
So I've just connected up the old milight gateway to test this.

[...]

PR #270, opened 2018-04-30
set cct bulb id correctly on on/off command

jstans:
ignore group id when setting bulb id in favour of defined group id against the on/off command code. discussed in #265


sidoh:
Thanks!

Think this should only happen for on/off.  There are not group-specific commands for brightness and temperature controls.

RGBW does the same thing:

https://github.com/jstans/esp8266_milight_hub/blob/3152a870bb5aedbba784fb165c10efdd0c738d10/lib/MiLight/RgbwPacketFormatter.cpp#L112-L115

sidoh:
Nevermind, I see what you're doing.  Thanks!

PR #271, opened 2018-04-30
Fix blank webpage on root (ERR_INCOMPLETE_CHUNKED_ENCODING)

### LINKED ISSUE #264: Webpage intermittently appears blank in 1.7.0-dev7 on Android Chrome
When this happens, the following seems to fix it (no idea why):

* Set "Request Desktop" to true
* Refresh
* Page now shows
* Can turn "Request Desktop" to false, page will still load

FrankZZ:
Regarding #264 blank webpage on root (ERR_INCOMPLETE_CHUNKED_ENCODING)

The content length was still set manually by the sendHeader call.

Also applied fix by sending an empty string just before closing the client socket:
https://www.esp8266.com/viewtopic.php?p=66764#p66875


FrankZZ:
Verified by performing curl, I no longer get `curl: (18) transfer closed with outstanding read data remaining`

sidoh:
Excellent.  Thank you!

I think this is what I meant to change with this commit https://github.com/sidoh/esp8266_milight_hub/commit/3c2e8169ccb366f2b5f8a4d26714d5fbf2634dc1, but obviously I missed. :)

PR #278, opened 2018-05-06
Fix #277: restart ESP after config portal times out

### LINKED ISSUE #277: Config-AP (ESPXXXXX) is always visible
Just noticed a strange thing: In my setup, router and miligh hub power cycles once a day. So the hub is back online before the router's wifi is back. The hub later connects fine to the wifi (commands are executes and my bulbs works normal) but the wifi "ESPXXXX" is still visible and can be connected to. Any ideas?

sidoh:


sidoh:
Restarting should fix.  The config portal should only appear if wifi is unavailable.

Are you positive the restart is working?

bombcheck:
Even with the fix, the portal wifi is visible.
To be clear: If the router wifi is available when the hub starts, the portal ap disappears. But when the router wifi is no there, the portal ap appears. The the router wifi comes up and the hub connects to it just fine. But then the config ap does not disappear.

sidoh:
@bombcheck think I figured it out.  

The config portal stays open for 3 minutes regardless of whether or not the AP becomes available during that time.

However, it always tries to connect to configured wifi at the end of that 3 minutes. 

So the `autoConnect` call was returning true, and the config AP was staying up.

https://github.com/sidoh/esp8266_milight_hub/pull/278/commits/f7f5d3fd7c64a31bc3bf7b7f5435f03da031b851 forces wifi into STA-only mode (i.e., turns off the AP).

Can you try it out with the latest code in this branch?

bombcheck:
Latest fix does the trick. Thx!

Issue #276, opened 2018-05-05
FUT038 Support


salex89:
Hi, I hope I'm not misusing the Issue feature here :-) . Does this solution support the FUT038 controller (it looks pretty similar to the FUT025 one). I've got some light strips and a couple of FUT038, so I'm seriously thinking in not buying any hub, but trying to make my own based on this, since I hope it will be better quality. So, just to know the state of it, has anybody tried the FT038?

As far as I see it, it is this [device](https://www.amazon.com/BTF-LIGHTING-Mi-Light-WiFi-Bridge-Controller/dp/B01H87DYR8/ref=sr_1_7?ie=UTF8&qid=1485715984&sr=8-7&keywords=milight) (the one you linked in the readme). It is without WiFi support, as in the product description, a hub is needed.  The WiFi hub is a separate device, maybe you wanted a link to it instead of this one?

Edit:
So I flashed the firmware onto a Wemos Lolin V3 board, everything went more or less straightforward, but I'm not sniffing anything for now. I'm using a NRF24L01+ ver A0 module. Now, It's possible I messed something up with the wiring or the module is defective (I have another one). I'll continue to investigate as far as my electronics/embedded knowledge goes :-) .

sidoh:
Yes, should work with RGB+CCT.  I've not tried it personally, but the product documentation says it works with the FUT092 remote, which is definitely supported.

> As far as I see it, it is this device (the one you linked in the readme). It is without WiFi support, as in the product description, a hub is needed. The WiFi hub is a separate device, maybe you wanted a link to it instead of this one?

Which link are you referring to?  This project is a replacement for the wifi hub.

> So I flashed the firmware onto a Wemos Lolin V3 board, everything went more or less straightforward, but I'm not sniffing anything for now. I'm using a NRF24L01+ ver A0 module. Now, It's possible I messed something up with the wiring or the module is defective (I have another one). I'll continue to investigate as far as my electronics/embedded knowledge goes :-) .

What device are you expecting to be able to sniff?  If you don't have a remote or a wifi gateway, sniffing isn't going to show anything.

To pair with a new device, follow these instructions:

https://github.com/sidoh/esp8266_milight_hub/wiki/Pairing-new-bulbs

salex89:
Looks like I overcomplicated my explanation :) .

[...]

PR #279, opened 2018-05-07
Extracting hue and saturation using rgbToHsl() instead of rgbToHsv()

salex89:
Saturation is not equal in these two models. There are slight differences which lead to inconsistencies if the same color is specified using both models.

### DIFF
--- a/lib/MiLight/MiLightClient.cpp
+++ b/lib/MiLight/MiLightClient.cpp
@@ -351,7 +351,7 @@ void MiLightClient::update(const JsonObject& request) {
     this->updateSaturation(request["saturation"]);
   }
 
-  // Convert RGB to HSV
+  // Convert RGB to HSL
   if (request.containsKey("color")) {
     JsonObject& color = request["color"];
 
@@ -366,12 +366,12 @@ void MiLightClient::update(const JsonObject& request) {
       && abs(r - b) < RGB_WHITE_THRESHOLD) {
         this->updateColorWhite();
     } else {
-      double hsv[3];
+      double hsl[3];
       RGBConverter converter;
-      converter.rgbToHsv(r, g, b, hsv);
+      converter.rgbToHsl(r, g, b, hsl);
 
-      uint16_t hue = round(hsv[0]*360);
-      uint8_t saturation = round(hsv[1]*100);
+      uint16_t hue = round(hsl[0]*360);
+      uint8_t saturation = round(hsl[1]*100);
 
       this->updateHue(hue);
       this->updateSaturation(saturation);


sidoh:
Thanks, @salex89.

Is there a reason to prefer one over the other?  Is the issue that the UI code uses HSV?

Think the UI actually uses HSV.

salex89:
Ah, sorry, it was late and I didn't give enough context to the PR nor connected it to the issue. Basically I described everything in this comment: https://github.com/sidoh/esp8266_milight_hub/issues/276#issuecomment-387179422.

To sum up, the saturation values in HSV and HSL are not equivalent. Sending the same color through the REST API in RGB and HSL does not yield the same light color. That's because the wrong converter method is used. When `rgbToHsl()` is used, you get the same color every time, no matter if the color is RGB or HSL coded. Extracting the saturation from HSV, when converting from RGB, and than using it in HSL is not correct.

As for the web client picker, I haven't used it that much, but that might be a separate issue worth looking in, and I probably will.

Should you accept this PR also depends on do you want to "break" existing RGB clients. With this fix they will get different colors than they previously had.

sidoh:
Thanks for clarifying.

I suppose I'm generally aware that saturation is a different measurement in HSV vs. HSL.

Am I wrong in thinking that it doesn't really matter which one we pick, as long as it's consistent?  I think HSV is currently used everywhere:

https://github.com/sidoh/esp8266_milight_hub/blob/03423159248bcd2b64659d1dd38d6eac83c93c71/lib/MiLightState/GroupState.cpp#L571
https://github.com/sidoh/esp8266_milight_hub/blob/03423159248bcd2b64659d1dd38d6eac83c93c71/lib/MiLight/MiLightClient.cpp#L371

the UI uses HSV to interpret state, but also casts to HSL in order to render a color in the browser:

https://github.com/sidoh/esp8266_milight_hub/blob/master/web/src/js/script.js#L597

What am I missing?  Is there something upstream of espMH that is assuming saturation is HSL?

salex89:
Well, yeah, you're more or less right. Consistency is key :-) . The only issue is the one I explained, that using same colors through the API specified using HSL and RGB does not yield the exact same color. But other than that no, HSL is not a requirement.

sidoh:
I guess I'm meaning to ask what is using HSL, because as far as I can tell, espMH is using HSV everywhere.

rjsachse:
Personally i think a shift to "FastLED" library would be a better go. Has a lot of features that could minimize some coding and open up to more features.

sidoh:
@rjsachse - want to better understand your suggestion.

How do you see FastLED relating to espMH?

I see espMH as a gateway that translates mostly sane APIs (REST, MQTT) into packets for milight's less sane proprietary 2.4 GHz protocol.

FastLED seems like an API that exposes a programming interface for different wire-based LED hardware.

rjsachse:
Using fastled you could adjust color temps or create new animations or even control a led device. I know it might be too much. It also has color conversations hsl hsv rgb hex etc

Just a idea

sidoh:
I guess I just don't see how espMH should interact with FastLED.  If you have a clear idea of what this would look like, happy to hear you out.

@salex89 - any thoughts on HSL vs. HSV?

salex89:
Sorry for the lack of activity, I was quite busy

I don't have any preference towards HSL or HSV, just to be clear. And I really do not know anything about the FastLED :-). And for the pull request, it can be closed, unless you want to continue the discussion, I was a bit confused by the API. It's clearer now, however:

- To be honest I would mostly like, when we are talking about the UI, for it be simple enough just to pick a color on a color wheel and light intensity and not care about what the software is using in the background.

- As for the API, one thing might be considered, and that is shutting down the color LEDs when the saturation is low, similar to the RGB case. When the saturation is low, the software does not turn the white LEDs off. For RGB, there is this check, so it does:

```
    if ( abs(r - g) < RGB_WHITE_THRESHOLD
      && abs(g - b) < RGB_WHITE_THRESHOLD
      && abs(r - b) < RGB_WHITE_THRESHOLD) {
        this->updateColorWhite();
```
The hue selection is fine, no issues with that.

PR #299, opened 2018-06-24
Use archive for limitlessled.com/dev as it's offline

Fantasmos:


sidoh:
Nice!

How are you imagining this should be consumed?  Might I suggest using [Github Pages](https://pages.github.com/) to host a static page?

Fantasmos:
Well, since from what I know they're limited to just one page each user it's not what I want to do. Right now I'm just hoping for people to download htmls as they need them.

Perhaps turning it to markdown could work?

Fantasmos:
To answer the previous question a bit more. I've converted the dev page to markdown, and broken it up a bit more into several fiels. If you have any suggestions they are appreciated!

Also, thanks a bunch for this project. Not only has it helped with another project i'm working on through remote emulation, but i'm hoping to finally get the V6 bulbs now that I can use them!

PR #305, opened 2018-07-17
Forward parameters in overloaded MiLightClient::prepare()

Gei0r:
The overloaded method shouldn't ignore `deviceId` and `groupId`.

This small programming omission is only noticable if someone uses `MiLightClient` directly; `MiLightHttpServer` doesn't use this overload.


sidoh:
Thank you!  Going to point this at the v1.8 branch.

PR #320, opened 2018-08-24
Fix UDP v6 RgbCct color skew

### LINKED ISSUE #319: UDP v6 RGB+CCT color offset
Migrating from the Mi-Light iBox I notice that my RGB+CCT colors set with the UDP v6 interface have changed (this is with Home Assistant).

The hue range 0x00-0xff seems shifted by 8 so that e.g. 0xb2 is pure blue with the iBox while espMH needs 0xaa for the same color. I think this is a bug because the UDP interface is supposed to be compatible?

Furthermore, the hue 0x00 should be pure red with espMH but that message is actually ignored and the bulb does not change color at all.

I have dumped my test messages below. Getting these with Home Assistant was a bit tricky because it currently has a similar issues where the hue is shifted by 2. I have removed the top of a bulb to verify that these are indeed the values where only 1 LED is on.

**iBox**
```
RED
bytearray(b'\x80\x00\x00\x00\x11\x19\x02\x00\x07\x001\x00\x00\x08\x01\x08\x08\x08\x08\x04\x00^')

GREEN
bytearray(b'\x80\x00\x00\x00\x11\x19\x02\x00\n\x001\x00\x00\x08\x01]]]]\x04\x00\xb2')

BLUE
bytearray(b'\x80\x00\x00\x00\x11\x19\x02\x00\r\x001\x00\x00\x08\x01\xb2\xb2\xb2\xb2\x04\x00\x06')
```

**espMH**
```
IGNORED
bytearray(b'\x80\x00\x00\x00\x11\x00\x11\x00\x10\x001\x00\x00\x08\x01\x00\x00\x00\x00\x01\x00;')

RED
bytearray(b'\x80\x00\x00\x00\x11\x00\x11\x00\x0c\x001\x00\x00\x08\x01\xff\xff\xff\xff\x01\x007')

GREEN
bytearray(b'\x80\x00\x00\x00\x11\x00\x11\x00\x14\x001\x00\x00\x08\x01UUUU\x01\x00\x8f')

BLUE
bytearray(b'\x80\x00\x00\x00\x11\x00\x11\x00\x18\x001\x00\x00\x08\x01\xaa\xaa\xaa\xaa\x01\x00\xe3')
```

amelchio:
This fixes #319 for me and UDP messages now set the same colors as my iBox.

I assume the call in the preset method would need to be similarly adjusted but I have no way of testing that at this time and I would like some feedback before spending more time on this.

### DIFF
--- a/lib/Udp/V6RgbCctCommandHandler.cpp
+++ b/lib/Udp/V6RgbCctCommandHandler.cpp
@@ -98,10 +98,6 @@ void V6RgbCctCommandHandler::handleUpdateColor(MiLightClient *client, uint32_t c
   for (int i = 3; i >= 0; i--) {
     const uint8_t argValue = (color >> (i*8)) & 0xFF;
 
-    if (argValue == 0) {
-      return;
-    }
-
-    client->updateColorRaw(argValue);
+    client->updateColorRaw(argValue + 0xF6);
   }
 }


sidoh:
thanks!  I believe I tested this with FUT015s more than a year ago.  I think I synced it on red (0x00).

Which bulbs are you testing with?

color 0x00 definitely shouldn't be ignored.  is that still happening?

amelchio:
I am testing on RGBWW and RGB+CCT bulbs that I believe are both called FUT014.

With this PR, 0x00 is no longer ignored.

sidoh:
sorry for the delay -- just want to get a chance to test it with my bulbs, but won't have time until probably this weekend.  thanks for your patience :)

sidoh:
and, of course, appreciate that you dug into this!

sidoh:
Confirmed this works for me and is right.  Not sure why the special case for 0x00 was there.

PR #333, opened 2018-09-09
Add NO_EXTRA_4K_HEAP build flag

jnordberg:
Had some issues building, not sure if this is the correct way to fix it but it got me going at least :)

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -27,7 +27,7 @@ lib_deps_external =
   ESP8266WebServer
 extra_scripts =
   pre:.build_web.py
-build_flags = !python .get_version.py -DMQTT_MAX_PACKET_SIZE=200 -DHTTP_UPLOAD_BUFLEN=128 -Idist -Ilib/DataStructures
+build_flags = !python .get_version.py -DMQTT_MAX_PACKET_SIZE=200 -DHTTP_UPLOAD_BUFLEN=128 -Idist -Ilib/DataStructures -D NO_EXTRA_4K_HEAP
 # -D DEBUG_PRINTF
 # -D MQTT_DEBUG
 # -D MILIGHT_UDP_DEBUG


jnordberg:
https://github.com/esp8266/Arduino/commit/e486887f18255a819c6a4e602752b646a058b965

sidoh:
Thanks, @jnordberg!

I've been investigating this issue as well.

The reason it's broken here is a little convoluted:

* Some cool features added by @cmidgley (Chris) required changes to the [WiFiManager](https://github.com/tzapu/WiFiManager) library
* Chris opened [a PR](https://github.com/tzapu/WiFiManager/pull/522) for the changes, but they weren't merged because it's already a feature in the dev branch or something.  Still hasn't been merged into master (~6 months later).
* Chris and I decided to point espMH at a fork of WiFiManager until the feature was available in master.
* The change to the Arduino SDK you found disables WPS by default, which was a breaking change for WiFiManager
* WiFiManager released a fix, but since we're operating on a fork they weren't pulled in

The route I was going down was to:

* Check if the feature Chris needed was available in WiFiManager yet (it's not, as far as I can tell).
* Since it wasn't, update our fork ([PR](https://github.com/cmidgley/WiFiManager))

Chris hasn't gotten back to me yet, but it probably makes sense for the fork to be tied to my account anyway.  So I set [sidoh/WiFiManager#cmidgley](https://github.com/sidoh/WiFiManager/tree/cmidgley) up.

This is working, as far as I can tell.

I think we don't want to add this build flag because it permanently steals 4KB of heap for the purpose of enabling WPS.  And I don't think anyone will want to use WPS because it's terrible.

sidoh:
Think we're all set, the build should be fixed in the v1.8.2 branch/release.

Gonna close, but please feel free to re-open if i'm missing something.  thanks again for investigating!

PR #339, opened 2018-09-18
Improve pinout documentation in README.md

saerdnaer:
see commit

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,11 @@ Both modules are SPI devices and should be connected to the standard SPI pins on
 
 [This guide](https://www.mysensors.org/build/connect_radio#nrf24l01+-&-esp8266) details how to connect an NRF24 to an ESP8266. I used GPIO 16 for CE and GPIO 15 for CSN instead. These can be configured later in the Web GUI, in Settings -> Setup.
 
-NodeMCU | Radio | Comment
+<img src="https://user-images.githubusercontent.com/40266/47967518-67556f00-e05e-11e8-857d-1173a9da955c.png" align="right" width="32%" /> <img src="https://user-images.githubusercontent.com/40266/47967520-691f3280-e05e-11e8-838a-83706df2edb0.png" align="left" width="22%" />
+
+<div width="40%">
+    
+NodeMCU | Radio | Color
 -- | -- | --
 GND | GND | Black
 3V3 | VCC | Red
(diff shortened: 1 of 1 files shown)


saerdnaer:
I'm not really sure, is that column should be dropped or not. When I clicked the link yesterday, the images did not load and my browser did not jump to the right place, for an unknown reason.  I only realised this after I already head connected the two modules without the photo. But it helped debugging my errors with photo afterwards (I switched MOSI and MISO). So we could rename the column from 'comment' to 'recommended color', if you want...

sidoh:
The colors are arbitrary, right?  This isn't some SPI convention is it?

If they're arbitrary, I think it's only helpful if accompanied with an image.  Otherwise it seems extraneous.  Maybe we can add an image?

saerdnaer:
Red and black are quite common for +5V and ground, the others are probably the default order when connecting a classic ribbon cable (aka "Row Dupont Cable") to a NRF24L01+:

image
(red to grey are still connected)

saerdnaer:
According to "Henrik Ekblad - Founder MySensors.org/OpenHardware.io" we can use https://www.mysensors.org/uploads/57c1a7461730a6c715b6b1b5/image/connect_nodemcu.png
https://www.mysensors.org/uploads/57c1a2411730a6c715b6b1b0/image/nrfTop.png 
– I will update the PR, when I have time for that. In which folder should we put the images? A new "docs" folder?

saerdnaer:
nrftop
connect_nodemcu

sidoh:
Could we also add a source for the images?  Other than that, LGTM.

sidoh:
@saerdnaer are you able to add sources for the images?  Will be happy to merge after that.  Apologize for the several rounds of back and forth.

saerdnaer:
What do you mean with "sources" the URLs are inside of https://github.com/sidoh/esp8266_milight_hub/pull/339#issuecomment-424005652 – please implemented the remaining changes yourself.

sidoh:
Sure.  I meant attribution.  Don't want the images linked to break, but they should be attributed to the person/website that created them.

sidoh:
Thank you!  Realized that since this PR has been open, the default pin for CE was changed from GPIO16 to GPIO4.  I'm going to merge this into a branch and make that edit before merging into master.

Appreciate the effort, and thanks for your patience with the super long iterations.

PR #367, opened 2019-01-01
Improve NRF24L01+ pin configuration

slootjes:
Because of the explicit link to this guide I kind of assumed the right pin was used by default which is not the case. The guide puts the CE pin on GPIO 4, not GPIO 16. Maybe it will save others some time if this was added to the readme.


slootjes:
I've committed the suggested changes. Changing the default to 4 would be nice so it matches the guide without requiring any changes. What was the reason to pick 16 as a default?

PR #373, opened 2019-01-05
Fix radio config endpoint

### LINKED ISSUE #372: Radio Configs JSON not listing all options
http://milight-hub/radio_configs only shows the following options: ["rgbw","cct","rgb_cct","rgb"].

However, the actual options are listed in the WEBUI correctly:
screen shot 2019-01-06 at 9 32 56 am

Missing the FUT809 and FUT091 options.

sidoh:
Fix #372


sidoh:
Fixed with this PR: #373.  Will be released as 1.8.7.

Note that I renamed the route /remote_configs to remain consistent with internal naming conventions.  But all remotes should be listed now.

Issue #418, opened 2019-03-15
Feature Request: MQTT Birth implementation


starwolf73:
Hi,
Using Home Assistant HASSIO and MQTT pluging, I'm trying to find a way to trigger when Milight_hub connect/reconnect to the MQTT server to assigned an automation task.  

Is it possible for the device to announce himself in a specific topic? Unless there is already a feature in MQTT Mosquitto I didn't saw.

Thanks
Eric

sidoh:
Hi Eric,

The 1.9.0 dev branches support MQTT LWT, which sends a message to a specified topic when the device disconnects.

There is not currently a way to send a message when first connecting, but that wouldn't be terribly hard to add.

starwolf73:
Thanks @sidoh. I believe the feature is call MQTT Birth. [https://www.home-assistant.io/docs/mqtt/birth_will/](url). I'll rename the title for sake of clarity.

sidoh:
This is released in [1.9.0-dev10](https://github.com/sidoh/esp8266_milight_hub/releases/tag/1.9.0-dev10).

starwolf73:
Thanks, I'll give a try this week and provide feedback.

rejoe2:
E.g. tasmota by default uses the LWT to inform about the online and offline status. I personally found this helpfull, but afai can see, there's no option to configure a "opposit to LWT" message. 

So is there an option to add also a online LWT message when the hub is connecting to the broker to revoke e.g. "offline"?

sidoh:
Yes, it's in the 1.9.0 branch, as mentioned above.  Try the RC release:

https://github.com/sidoh/esp8266_milight_hub/releases/tag/1.9.0-rc2

rejoe2:
My remark already refferenced to version 1.9.0-rc2. Both message types (LWT and birth messages) are great! But one has to "clean up" the LWT message "manually" on the HA controller side (may be different in HASS), as there is only one message to the LWT topic, but not a second, indicating the hub is online again. This one (birth) is to a different topic and in a different structure - single word payload (or whatever is configured) vs. JSON. 
So my suggestion would be to also send a second "birth" message info to LWT also (same way as eg. tasmota devices do).

sidoh:
[...]

PR #435, opened 2019-04-15
Use single topic for MQTT client status (LWT, birth, etc.)

sidoh:
As suggested/discussed in this comment thread: https://github.com/sidoh/esp8266_milight_hub/issues/418#issuecomment-482675232

Thanks, @rejoe2!


odis68:
Tell me how to useLWT. Because messages must be 2. The first is entered into the broker when connected with an indication of the topic, and the second is periodically sent to the broker for a status condition. Is it possible to use a separate topic (milight/states/LWT) for LWT or should it be combined with any of the other topics?

odis68:
image

rejoe2:
Are you using latest version (1.9.0-rc3)?

odis68:
Sorry for an example from another project. This is not disrespectful but for example only.
image

Yes, I have installed latest version (1.9.0-rc3)

sidoh:
@odis68, the name of the setting changed.  It's now `mqtt_client_status_topic`.  You'll need to reconfigure it.

sidoh:
Guessing the data on your `status/LWT` topic is a left over retained message.

odis68:
image

sidoh:
That is not RC3.  Looks like RC2 or earlier.  RC3 here:

https://github.com/sidoh/esp8266_milight_hub/releases/tag/1.9.0-rc3

odis68:
r. RC3 
I cleared the topic () by entering an empty value but after resetting the same value.

sidoh:
The UI is definitely from RC2.

Can you maybe try force-refreshing?  Perhaps the UI got cached by the browser or something.

odis68:
You were right!!! The problem in the browser helped clear the cache !! Thank you very much you have fulfilled several trials at once.

PR #455, opened 2019-05-06
Temp/Humidity Sensors

jmw6773:
Sidoh,
  Is there any interest in adding support for basic environment sensors for temperature/humidity? (DHT, BME280, SHT21, etc)
  I know it doesn't add anything useful to the core purpose of controlling MiLights, but I have my hub in the attic where it gets hot in the summer and wanted to monitor the temperature without adding another ESP just for this purpose.

  I added UI and Setting elements to:
- enable/disable this functionality
- select which sensor to use
- enter the DHT pin (DHT sensors)
- select the I2C address (BME280, SHT21)
- select whether to report C of F values
- what MQTT topic to publish onto
- how often to update the sensor/publish

  Unfortunately I've only tested with the DHT22 and BME280 sensors. I've ordered the SHT21 sensor, but it is currently untested.

  Adding support for these sensors adds about 10kb onto the compiled firmware size. I'm not sure how much RAM it uses. I've been running this firmware without issues for a week.


sidoh:
Very nice!  Also super complete.

I totally understand that this is useful for your situation, and I'm super impressed with the thoroughness of your PR.  Looks really good.

That said, I don't think we should merge this.  Like you said, this project has a pretty clear and specific purpose, and I'm very hesitant to stray away from that.

I'd be happy to point others looking for this functionality your way.

btw - there are some changes coming down the pike that'll cause some minor conflicts.  Mainly, I'm updating ArduinoJson from v5 to v6.  Example update here:

https://github.com/sidoh/esp8266_milight_hub/compare/rich_web_server#diff-61634c72f47505960e765c08c0eb5218L172

Issue #113, opened 2017-09-15
Support for transitions


corneyl:
First, thanks for this package, works perfect!

I couln't find any mentions of this before, but would it be possible to support transitions? 
[`mqtt_json`](https://home-assistant.io/components/light.mqtt_json/) in Home Assistant already sends this parameter in the MQTT messages, but I do not know whether this is feasible to implement.
In my opinion this should be implemented in the bridge, and not in Home Assistant, as timing can be done a lot more precise on a ESP (realtime vs OS), and MQTT messages have more overhead opposed to RF messages.

sidoh:
Just to make sure I'm on the same page: a "transition" just means that you slowly fade from the previous state state to the new state?

I agree that it makes sense for that to be implemented by the controller rather than HASS or whatever else.

That is blocked by esp8266_milight_hub being stateful. It currently has no memory for states -- it just sends commands as it gets them, and doesn't keep track of what state bulbs are in. 
Statefulness will add a lot of complexity, and will take a fair amount of effort to implement, but is a blocker for several features that people care about (eg #68).

sidoh:
Oh, does the transition command include both the original state and the desired state? That wasn't clear to me. If it does, then you're right. Not strictly necessary to keep track of bulb state.

corneyl:
Ok, sorry, didn't think of that. 
I don't see anything mentioned in the docs about sending the original state so I don't think it's included. So then it wouldn't be possible to create transitions without making the whole thing stateful.

sidoh:
Cool, makes sense. Thanks for confirming!

corneyl:
It looks like v1.6 will be stateful, does that mean that transitions will be possible in the future? That would be awesome! 👍

sidoh:
Yes, will enable transitions.  Transitions will probably not be in v1.6, but this is probably about half of the hard part.  Other hard part is having some kind of pipeline for queued work.

sidoh:
Should say that this sounds like a fun feature to work on, so it'll probably happen at some point :)

bhaap:
As a easy feature to add would be to make the OFF command send brightness to 0 wait for X ms depending on current brightness and then send OFF.

And ON command send ON then Brightness 255 if no other brightness was given.

Easy way to get a smooth on/off atleast.

The only downside to transitions by the "hub" is that we get alot of radiotraffic, say swapping from one color to another. That would take like atleast 100 radio commands to get a smooth transition. 

[...]

PR #479, opened 2019-06-16
New feature: queued packet sends

sidoh:
This new feature adds a queue for packet sends, which allows the various integration points (REST, MQTT, etc.) to enqueue command packets instead of synchronously sending them.  

The observable result to the user is for MQTT and UDP, vastly improved performance.  Network commands are not blocked on sending the RF packets.  They simply enqueue the commands they want sent and yield back to the system.  In testing, I was able to send MQTT commands using `mosquitto_pub` with absolutely no throttling and there was no observable performance impact.

For REST, there is currently no difference.  PUT/POST state still handle the command synchronously.  If desired, this is easy to patch.

This change also enables new features like #113.


coehn1:
Hi,
Great work!
- For REST, there is currently no difference. PUT/POST state still handle the command synchronously. If desired, this is easy to patch.
Could you please implement this?

sidoh:
Yeah, I can look into it.  I actually added special code to keep the REST API synchronous.  `PUT` returns the state of the group after patching, but state is only updated after packets are sent.  So it was returning invalid results.

I think the options are:

* Make sync/async a setting
* Make the API always asynchronous, and add a route to query the queue status
* Make `GET` wait for sends to complete, update usages that rely on state `PUT` to instead `GET` after `PUT`

I'll think about which of these make the most sense.  Let me know if you have opinions.

PR #488, opened 2019-07-20
New feature: transitions

sidoh:
## Summary of changes:

* Support for transitions (details follow)
* A bunch of miscellaneous cleanups and improvements
* A few bug fixes

## Transitions

This is a weighty feature that enables transitions between states.  For example, you can fade from 100% brightness to 0%, or between two colors.

This can be accomplished either by:

* One of the newly added REST route `/transitions` (details documented in README)
* By adding the `{"transition":x}` key to the state command.  `x` is some the number of seconds you want in-between.

## Bug fixes

* Linked List library had a bug that prevented arbitrary nodes from being deleted
* Use PROGMEM in a couple of new places.  Will marginally improve free heap.
* Fix a test that was broken in 1.10.
* Fix a memory leak that occurred when settings were saved and there is a non-empty state cache.

## Improvements

* Pull string constants like `"brightness"` into actual constants rather than having the literals scattered throughout the codebase.  Prevents typo bugs and decreases heap/flash utilization.
* Refactor main JSON command -> milight packets handler to not use a mess of conditionals.

### DIFF
--- a/lib/DataStructures/LinkedList.h
+++ b/lib/DataStructures/LinkedList.h
@@ -194,6 +194,7 @@ bool LinkedList<T>::add(T _t){
   if(root){
     // Already have elements inserted
     last->next = tmp;
+    tmp->prev = last;
     last = tmp;
   }else{
     // First element being inserted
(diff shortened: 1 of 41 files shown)


sidoh:
Reference issue: #113

PR #510, opened 2019-08-30
Fix broken build with latest Arduino libs - bump Websockets to ~2.2.0

pinkeen:
There was a compilation error when building the external Websockets dependency with latest Arduino libraries (2.5.X).
It was caused by SSL code changes in Arduino and since has been fixed in Websockets 2.2.0.

See Links2004/arduinoWebSockets#447 for details.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -20,7 +20,7 @@ lib_deps_external =
   ArduinoJson@~6.10.1
   PubSubClient@~2.7
   ratkins/RGBConverter@07010f2
-  WebSockets@~2.1.2
+  WebSockets@~2.2.0
   CircularBuffer@~1.2.0
   PathVariableHandlers@~2.0.0
   RichHttpServer@~2.0.2


sidoh:
Hiya @pinkeen, thanks for looking into this.  I actually worked around this in the v1.10 branch by [setting the `USING_AXTLS` constant](https://github.com/sidoh/esp8266_milight_hub/blob/v1.10.0-wip/platformio.ini#L33-L34).

Obviously it's nicer to not need to set this flag, so we'll do this.  Thanks again!

(unimportant: funnily enough, I actually opened the issue you linked 😄)

pinkeen:
Didn't notice that it's your issue, quite an irony :)

Thank you, and keep up the good work! Everything worked out of the box for me (well, except sniffing, but I suppose I need to take a closer look at my remote and see if I can rev-engineer it).

BTW I am new to platform.io, is it possible to lock the version of platform libraries (I am thinking Arduino libs for ESP)?

PS I am just recently trying to get back to IoT, last time I was doing something here was not long after the first Arduino's came out - the possibilities now are astonishing :)

sidoh:
> BTW I am new to platform.io, is it possible to lock the version of platform libraries/framework (I am thinking Arduino libs for ESP)?

Yes.  I'm already doing it here: 

https://github.com/sidoh/esp8266_milight_hub/blob/master/platformio.ini#L13

PlatformIO's platform versions aren't correlated to the ESP8266 SDK versions, but 1.8 corresponds to 2.4.

The issue here wasn't quite what you said in the first comment.  It's actually that a new version of WebSockets (2.1.4) became incompatible with older versions of the SDK (2.4 and lower).

I tried to upgrade to 2.5 a few months ago and several users reported issues.  At the time, it seemed too unstable to be worthwhile.  So I downgraded back to 2.4.

PR #534, opened 2019-09-30
Added file upload accept *.bin, *.json

rjsachse:
Only accept *.bin for firmware and *.json for settings restore


sidoh:
Sounds good.  Are you able to run the web portion of the build which will re-generate the progmem artifact?  It's built into the platformio build, but it's allowed to fail silently.  If it is failing, it's probably because you need to install `node` and `npm`.

PR #536, opened 2019-10-03
Added favicon and Bootswatch theme support

rjsachse:
i designed it on favicon.cc
made it look similar to milight app icon
just a hue wheel, not sure if it works on IE but does on Chrome

This is what it looks like favicon

 also added Bootswatch theme support

also changed some design elements to make it suit better to mobile


sidoh:
Thanks very much for the PR!  Love all of this.

Comments:

- The icon looks great.
- I like the rearranging you did.  This looks better.
- Themes are a nifty idea, and I like the way you've implemented it.  Some of the components I'm using don't seem to play terribly well with many of the themes, though.  e.g., darkly looks like this: \
  

Not sure if there's an easy fix, but don't love the idea of shipping themes that don't really work functionally.

web/src/js/theme-select.js
```diff
@@ -0,0 +1,67 @@
...
+    var btnCssHeight = $("#btnHeight").css("height");
+    var raw = btnCssHeight.replace("px", "")
+    $(".plus-minus-group .title").css("height", btnCssHeight);
+    $('input[name="status"]')
+      .bootstrapToggle('destroy')
+      .bootstrapToggle({
+        height: raw
+      });
+  }, 1000)
+}
```
rjsachse: i tried to set the mode buttons to change height depending on theme after 1 sec, had some issue while developer mode open not setting till refresh, the issue is i haven't found out a way to tell when the css has finished loading so there is a delay to set the height, we could increase the delay.

web/src/index.html
```diff
@@ -43,6 +44,12 @@
       </div>
       <div class="collapse navbar-collapse">
         <ul class="nav navbar-nav">
+          <li class="dropdown" id="theme-dropdown">
+            <a id="theme-label" href="#" class="dropdown-toggle" data-toggle="dropdown">Theme<b class="caret"></b></a>
+            <ul id="theme-list" class="dropdown-menu">
+              <li><a href="#" class="change-style-menu-item">Bootstrap (Default)</a></li>
+            </ul>
+          </li>
```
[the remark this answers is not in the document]
rjsachse: i wasn't sure where to put it, but yes admin is good, i can fix that

rjsachse:
the other issue with the buttons not showing right is a easy fix too, the issue is the the btn has secondary colour should which is a bootstrap 4 colour but we use bootstrap 3 so should to something like  btn-default

web/src/js/theme-select.js
rjsachse: i think i worked it out

rjsachse:
I am closing this because i had my branches wrong, still getting use to git. created new pull request with each new feature/modification

PR #573, opened 2019-12-19
Update README.md

h4nc:
Testet connecting an LED at the same pin as the internal LED. The external LED will always be inverted against the internal LED. 
So if the internal is off the external is on. So one LED will always be on. This means it's better to use another pin (eg. GPIO5).
So let's go back, sorry. We still changed the example pin, which was connected to CE before.


tbrasser:
If it's inverted, you can reverse the led and connect it to Vcc right?

h4nc:
You are totally right, just tried it and it works. Will add that to the Readme. Thanks a lot!

PR #583, opened 2020-01-04
added support for Casalux cct panel

aeinstein:
added support for the Casalux cct panel sold from Aldi in Germany. Please merge only commit #
db77d0e. This is my first pull request, and I can't find out how to only select this commit for pull, sorry

### DIFF
--- a/lib/DataStructures/LinkedList.h
+++ b/lib/DataStructures/LinkedList.h
@@ -152,11 +152,10 @@ ListNode<T>* LinkedList<T>::getNode(int index){
 
   while(_pos < index && current){
     current = current->next;
-
     _pos++;
   }
 
-  return false;
+  return current; //might be null
 }
 
 template<typename T>
@@ -177,10 +176,12 @@ bool LinkedList<T>::add(int index, T _t){
          *_prev = getNode(index-1);
   tmp->data = _t;
   tmp->next = _prev->next;
+  //we know that there is a next node, since 0<index<_size
+  tmp->next->prev = tmp;
+  tmp->prev = _prev;
   _prev->next = tmp;
 
   _size++;
-
   return true;
 }
 
@@ -267,6 +268,7 @@ T LinkedList<T>::shift(){
     ListNode<T> *_next = root->next;
     T ret = root->data;
     delete(root);
+    _next->prev = NULL;
     root = _next;
     _size --;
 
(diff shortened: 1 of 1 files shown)


normen:
@aeinstein: In case you didn't find out yet: Normally you would make a branch in your repository to do your development work and then just merge the wanted changes to your master and offer that for the pull request.

If you already changed master you can do it the other way around and make a branch with only the wanted changes and offer that for PR.

Merging a single commit from PR like this is not really possible.

Cheers!

sidoh:
Totally missed this, sorry @aeinstein.

Appreciate you adding this, and impressed that you were able to make this deep of a change unaided.

Couple of high level concerns I have with merging this --

* Correct me if I'm wrong, but this seems like a niche device.  Adding new remote types and especially radio configs introduces overhead.  So obviously I'm hesitant to do that unless a lot of people use the device.
* Seems like there are some extraneous changes here.  Namely changing the `RF24Channel` enum to `uint8_t` (I prefer enums), some references to an `NLG` remote type that doesn't look to be fully supported in your changes, and some changes to the default settings.

Let me know what you think.

aeinstein:
@sidoh : I totally agree with you, nobody wants more overhead. I think about some checkboxes in the config tab for enable/disable the types of radios (channels, syncwords) that are used by the user. (I only have casalux and NLG)

Also I dont wanted the other changes (free channel input) to be merged, this was only for understanding and testing purposes. 
Sorry for my wrong usage of github, I used to act with my own git server. (blame on me).
I suggest to cancel the merge request and I make a new branch for the casalux impl. and one more for the changes of radio config.

The NLG impl. is not yet ready and I struggle with it. Cant find the Checksum Algo. Perhaps you can help, if I give you the sniffed raw data?

Eisenhauer1987:
@aeinstein 

I have a casalux panel, too. But I cannot compile your version successful. I get several error for nodemcuv2, example:

`fatal: not a git repository (or any of the parent directories): .git`

Can you help me? Compiling of the normal version runs fine

shamasis:
@aeinstein ... I am a general user of this software and I learnt a lot from your PR (as to how I can go about submitting fixes and features.)

If I may, I think for this PR to get merged it needs the following:
1. Resend a fresh clean PR with only the changes that needs to be merged
2. Have the commits be atomic and build up to the end state with relevant commit messages
3. If anything is changed, accompany comments or documents explaining why.

Lastly, various device type support can be made an early stage config read and then flagging via and enum to avoid commuting unneeded code-paths and not have overheads.

gtawelt:
I have a NLG Panel and cant get to sniff the codes

aeinstein:
> I have a NLG Panel and cant get to sniff the codes
This module don't understand NLG, cause I dont understand the checksum. I suffered to find the right algorhythm, sorry. I will try it again, when I rebuild my smarthome

PR #659, opened 2020-10-12
Fix pio errors

rjsachse:
changed lib depends,
git ignore dist files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@
 /web/package-lock.json
 /dist/*.bin
 /dist/docs
-/dist/*.h
 .vscode/
 .vscode/.browse.c_cpp.db*
 .vscode/c_cpp_properties.json
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 /web/package-lock.json
 /dist/*.bin
 /dist/docs
+/dist/*.h
 .vscode/
 .vscode/.browse.c_cpp.db*
 .vscode/c_cpp_properties.json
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -9,6 +9,7 @@
 /web/package-lock.json
 /dist/*.bin
 /dist/docs
+/dist/*.h
```
sidoh, line 12: This should actually not be in `.gitignore`. It's a little ugly, but I intend to check in the compiled webpage so that people don't need to have a webdev environment set up in order to build the project (the build script falls back on using the existing `.h`).

platformio.ini
```diff
@@ -15,11 +15,11 @@ board_f_cpu = 160000000L
 lib_deps_builtin =
   SPI
 lib_deps_external =
-  sidoh/WiFiManager#cmidgley
+  WiFiManager=https://github.com/sidoh/WiFiManager.git#cmidgley
   RF24@~1.3.2
   ArduinoJson@~6.10.1
   PubSubClient@~2.7
-  ratkins/RGBConverter@07010f2
+  https://github.com/ratkins/RGBConverter.git
```
sidoh, line 22: Can we fix this to the same commit? Hoping the same syntax works `@07010f2`. I prefer to have all dependencies locked to a particular version.

sidoh (commented):
tyvm for looking into this! left a few comments, but looks great.

rjsachse:
makes sense with index.h, added version control to rgb converter now # not @

sidoh:
👍  beautiful! thank you.

PR #666, opened 2020-11-02
Fix brightness setting command

PaulSzymanski:
I am using espMH with a cct led controller (fut035) and a remote (fut007). I have set Home Assistant to forward messages from the mqtt_update_topic (of the remote) to the mqtt_topic for the emulated cct remote. Set up like this I was able to control the light via the fut007, except the brightness control was not working.
I found that sending `level_up/down` instead of `brightness_up/down` was working.
With this commit my setup is working as expected.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -526,7 +526,7 @@ components:
 
         * `night_mode`.  Most devices support a "night mode," which has LEDs turned to a very dim setting -- lower than brightness 0.
 
-        * `level_up`. Turns down the brightness. Not all dimmable bulbs support this command.
+        * `level_up`. Turns up the brightness. Not all dimmable bulbs support this command.
 
         * `level_down`. Turns down the brightness. Not all dimmable bulbs support this command.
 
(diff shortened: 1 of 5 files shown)


PaulSzymanski:
I believe the RgbPacketFormatter and FUT020PacketFormatter need to be adjusted as well?
https://github.com/sidoh/esp8266_milight_hub/blob/master/lib/MiLight/FUT020PacketFormatter.cpp#L63-L67
https://github.com/sidoh/esp8266_milight_hub/blob/master/lib/MiLight/RgbPacketFormatter.cpp#L113-L115

sidoh:
Good catch, @PaulSzymanski! Really appreciate you digging into it.

Yes, those look like the FUT020 and Rgb packet formatters need to be changed as well. I have a bit of time today so planning on knocking this out. If you have a chance to make the changes yourself, I'll get those merged in. Otherwise I can make them on my end and be sure to give you credit =)

PaulSzymanski:
There you go. I did not test if it works with these remotes, tho!

As a side note: This is probably a breaking change as people will already have their environments set up to work with `brightness_up/down`? I know that this would be a breaking change for my setup.

sidoh:
Yes, good call.

Just to make sure I'm following what's going on here... the issue is that:

* Messages published to the `mqtt_update_topic` use `level_up` and `level_down` commands
* The hub only responds to `brightness_up` and `brightness_down` commands

If that's accurate, the backwards compatible change which I think would also fix the issue would be to have the hub respond to both `brightness_` and `level_` up/down commands. Let's do that instead.

PaulSzymanski:
I think its the other way around.

- Messages published to the `mqtt_update_topic` use `brightness_up` and `brightness_down` commands
- The hub only responds to `level_up` and `level_down` commands

#668 should be a non breaking fix

sidoh:
Gonna close this in favor of #668. Thanks for your efforts & patience with the back and forth @PaulSzymanski :)

PR #715, opened 2021-02-25
Fix color transition memory leak

MunkeyBalls:
Fix for a memory leak which could occur when a light does a color transition to itself.

Example of a transition being stuck:
```
{
  "transitions": [
    {
      "id": 1520,
      "period": 4294967295,
      "last_sent": 0,
      "bulb": {
        "device_id": 1,
        "group_id": 1,
        "device_type": "rgb_cct"
      },
      "type": "color",
      "current_color": [
        255,
        167,
        91
      ],
      "end_color": [
        255,
        167,
        91
      ],
      "step_sizes": [
        0,
        0,
        0
      ]
    },
}
````
 #600 #523

### DIFF
--- a/lib/Transitions/ColorTransition.cpp
+++ b/lib/Transitions/ColorTransition.cpp
@@ -133,7 +133,7 @@ void ColorTransition::step() {
 }
 
 bool ColorTransition::isFinished() {
-  return finished;
+  return currentColor == endColor; 
 }
 
 void ColorTransition::childSerialize(JsonObject& json) {
--- a/lib/Transitions/ColorTransition.cpp
+++ b/lib/Transitions/ColorTransition.cpp
@@ -73,7 +73,6 @@ ColorTransition::ColorTransition(
   , stepSizes(stepSizes)
   , lastHue(400)         // use impossible values to force a packet send
   , lastSaturation(200)
-  , finished(false)
 {
   int16_t dr = endColor.r - startColor.r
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by MunkeyBalls** [comment]
Still testing, but first impressions are good. No transitions stuck yet after a few hours of use. Free-heap seems stable as well.

**Comment by sidoh** [comment]
Sweet! Can we remove `finished` all together?

Think this would be updating the condition to just check `if (isFinished())` and removing `finished` from the header and constructor.

File: lib/Transitions/ColorTransition.cpp [hunk 1]
```diff
@@ -122,10 +121,9 @@ void ColorTransition::step() {
...
     lastSaturation = parsedColor.saturation;
   }
-
-  if (currentColor == endColor) {
-    finished = true;
-  } else {
+  
+  if(!isFinished())
+  {
     Transition::stepValue(currentColor.r, endColor.r, stepSizes.r);
```
sidoh, line 127:
```suggestion
  
  if (!isFinished()) {
    Transition::stepValue(currentColor.r, endColor.r, stepSizes.r);
```

Just to keep styling consistent :)

sidoh:
Looks great! just the styling comment and should be good to merge.

Thanks for looking into this!

MunkeyBalls:
No worries, happy to help and this means I can finally run the latest version :)

Issue #714, opened 2021-02-23
Long press is almost always activates night mode (unsupported button/device?)


golddragon007:
Hello,

I don't have Mi light devices I have only the remotes and I write automation to control WLED devices with them, so I don't know if this is the correct behaviour or not, but I have the following Mi Light 8 zone device (identified as FUT089):
![image](https://user-images.githubusercontent.com/8385152/108906339-126c0700-7621-11eb-9636-491945d188f9.png)
Where if I long press any of the following: W (white mode), ON, OFF, S- (mode speed down), S+ (mode speed up) then I will receive night mode, which is really strange for the last 3 buttons. From the Command/Argument stand of point in the packet, they are all different. If the original device doesn't go into night mode in any of those cases, is it possible to remove it from the MQTT message and replace it with something else?

Another interesting case would be for this use case if I would be able to overwrite the command name (in MQTT) from the UI per controller and then I can use it for different function the long presses of those buttons.

Then I have the following device, which is identified as FUT091:
![image](https://user-images.githubusercontent.com/8385152/108908454-9cb56a80-7623-11eb-82e4-3d352aa3c753.png)
Here similar happens, any on/off button that I press long will send night mode. But here there's MQTT message randomness also. Let's say I long-press the first time the zone 1 on or off button, and I will get the night mode command on zone 0. Then I long press on zone 2 and I get it in zone 1 the command, but only the long press. If after these, I long-press zone 4 I get it in zone 2 the command... Then to make it more confusing, sometimes (in rare cases) I get into the correct zone the command. Also sometimes I receive in this case messages to the milight/(updates OR states)/0x0/rgb_cct/1 topic.

sidoh:
The packet handler is here:

https://github.com/sidoh/esp8266_milight_hub/blob/master/lib/MiLight/FUT089PacketFormatter.cpp

Might be because night mode is prioritized above speed up/down. 

I unfortunately don't have a lot of free time for this project right now. I'd encourage you to try playing with this packet parser if you think something is amiss. I'd happily review a PR.

golddragon007:
On the original bulbs, which button long-press sets the device to night mode?

sidoh:
I think long-pressing the "ON" button sends the bulbs into night mode. I have a physical FUT089 and some FUT015s to test with, but won't be able to dig them out until the weekend.

golddragon007:
[...]

PR #718, opened 2021-03-05
Don't activate every longpress nightmode

golddragon007:
Partially covers #714 .
It will send only night mode on long OFF button press.
It will send on the update channel "long_press" with true/false when it's possible to do it (ON, OFF, S-, M, S+, W).

### DIFF
--- a/lib/MiLight/FUT089PacketFormatter.cpp
+++ b/lib/MiLight/FUT089PacketFormatter.cpp
@@ -111,22 +111,30 @@ BulbId FUT089PacketFormatter::parsePacket(const uint8_t *packet, JsonObject resu
 
   uint8_t command = (packetCopy[V2_COMMAND_INDEX] & 0x7F);
   uint8_t arg = packetCopy[V2_ARGUMENT_INDEX];
+  bool long_press = (packetCopy[V2_COMMAND_INDEX] & 0x80) == 0x80;
 
   if (command == FUT089_ON) {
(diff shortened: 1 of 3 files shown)


lib/MiLight/FUT089PacketFormatter.cpp
```diff
@@ -111,22 +111,30 @@ BulbId FUT089PacketFormatter::parsePacket(const uint8_t *packet, JsonObject resu
...
     } else if (arg <= 8) { // Group is not reliably encoded in group byte. Extract from arg byte
       result[GroupStateFieldNames::STATE] = "ON";
       bulbId.groupId = arg;
+      result[GroupStateFieldNames::LONG_PRESS] = long_press;
     } else if (arg >= 9 && arg <= 17) {
       result[GroupStateFieldNames::STATE] = "OFF";
       bulbId.groupId = arg-9;
+      result[GroupStateFieldNames::LONG_PRESS] = long_press;
+      if (long_press) {
+        result[GroupStateFieldNames::STATE] = "ON";
```
Jojo-1000, line 135: This changes the behavior of the lights. Now when I keep pressing the off button, it will show one "OFF" state and then "ON" with night_mode. Before it was continuous "OFF" with night mode. I am not sure which way is better, but to me it is confusing that I am pressing off but the command shows as on.

golddragon007:
Uhh, I've completly forget about this... was this fixed meanwhile?

PR #725, opened 2021-05-09
Esp32

Linkenelis:
TODO erase config, to delete the wifi credentials
just put:
          WiFi.disconnect(true, true);
          delay(1000);
in the ESP32 part of MiLightHttpServer.cpp 

  if (requestBody.containsKey(GroupStateFieldNames::COMMAND)) {
    if (requestBody[GroupStateFieldNames::COMMAND] == "restart") {
      Serial.println(F("Restarting..."));
      server.send_P(200, TEXT_PLAIN, PSTR("true"));

      delay(100);

      ESP.restart();

      handled = true;
    } else if (requestBody[GroupStateFieldNames::COMMAND] == "clear_wifi_config") {
        Serial.println(F("Resetting Wifi and then Restarting..."));
        server.send_P(200, TEXT_PLAIN, PSTR("true"));

        delay(100);
        #ifdef ESP8266
          ESP.eraseConfig();
        #elif ESP32
          Serial.println("wifi reset...");
          WiFi.disconnect(true, true);
          delay(1000);
        // TODO erase config
        #endif
        delay(100);
        ESP.restart();

        handled = true;
      }
  }


grote-beer:
Using this i noticed an exception: "[E][vfs_api.cpp:27] open(): group_states/1000200 does not start with /" during startup and during operation.

Changing the FILE_PREFIX to include the leading '/' in /lib/MiLightState/GroupStatePersistence.cpp resolves this issue.

butteryak:
Checking in on this, was the ESP32 version ever finished? is there a working firmware? tnx++

Linkenelis:
> Checking in on this, was the ESP32 version ever finished? is there a working firmware? tnx++
Look here for the source code: https://github.com/Linkenelis/esp8266_milight_hub/tree/esp32

It works on ESP32.
I tested on a TTGO-T7-mini32 everything works exept the WiFi mode setting
Not a finished project and not maintained

To combine it with esp8266 based on environment is a lot of work. Mainly due to the differences in WiFi. The WiFi mode setting on the webserver does not work for ESP32 (different hardware), so you should have a different webpage too (not done here). Without the WiFi mode setting I think it can be done quite easily.

second thing is: what to do with the enormous amount of extra processor time...Display is nice? But maintaining it? Or automate the display generation according to setup of devices would be nice. Not my cup of tea :-)

So, no benefit to use ESP32...

1ucay:
Hi, what about WT32-ETH01 ESP32, which has ethernet module. With this it will be great solution.

ristomatti:
There's also a more recent fork of WiFiManager library available which supports both ESP8266, ESP32: https://github.com/tzapu/WiFiManager. I doubt it supports ethernet though which would be a welcome improvement. I've been using the current version of ESP8266 Milight Hub at two locations for years now with very little issues. The issues have so far always been WiFi related since the fix has required either resetting the ESP or rebooting my WiFi router.

karol27:
I need it working with WT32-ETH01. Is it possible? In my situation connection with ethernet is only by wire.

PR #736, opened 2021-10-04
Added "min_mireds" & "max_mireds" to the HA discovery payload

jvanderzande:
Added "min_mireds" & "max_mireds" to the HA discovery payload so the system know the scale is not the standard 150-500 range and adjust the scale.

### DIFF
--- a/lib/MQTT/HomeAssistantDiscoveryClient.cpp
+++ b/lib/MQTT/HomeAssistantDiscoveryClient.cpp
@@ -1,5 +1,6 @@
 #include <HomeAssistantDiscoveryClient.h>
 #include <MiLightCommands.h>
+#include <..\Helpers\Units.h>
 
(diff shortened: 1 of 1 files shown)


jvanderzande:
@sidoh ,  Would it make more sense to move the 2 lines I added within this section as that is only for the CCT bulb, which the min/max_mireds is defined for?   Like this: 
```
  // These bulbs support adjustable white values
  switch (bulbId.deviceType) {
    case REMOTE_TYPE_CCT:
    case REMOTE_TYPE_FUT089:
    case REMOTE_TYPE_FUT091:
    case REMOTE_TYPE_RGB_CCT:
      config[GroupStateFieldNames::COLOR_TEMP] = true;
      config[F("max_mireds")] = COLOR_TEMP_MAX_MIREDS;
      config[F("min_mireds")] = COLOR_TEMP_MIN_MIREDS;
      break;
    default:
      break; //nothing
  }
```
I can make the PR for this change when you want. :)

sidoh:
Sounds right, yea. Thanks!

PR #738, opened 2021-10-07
Moved mireds into cct section

jvanderzande:
Moved HA dicovery mireds fields into CCT  section so they will only be set when an bulb is involved supporting CW-WW.


sidoh:
Ah yea, I need to tag a release. Assuming CI is still working it should publish it.

Issue #792, opened 2023-04-30
Home Assistant MQTT format now requires color_mode to be sent


VossenTech:
### Describe the bug

Updates to home assistant mqtt light json format require an additional parameter to be sent in the json.
Parameter: "color_mode"
Possible values: "onoff", "brightness", "color_temp", "hs", "xy", "rgb", "rgbw", "rgbww", "white"

Without sending this parameter  the entities no longer have brightness, color_temp, or RGB attributes due to the color_mode being unknown. The lights are still fully functional using manual control however any automations that utilize any of these attributes as triggers or checks the current values are broken.

I did some quick testing and it seems to me that when milight hub is sending "bulb_mode":"color" HA wants it to also be sending "color_mode":"rgb", and when its "bulb_mode":"white" than HA wants "color_mode":"color_temp". Hopefully that makes it an easy thing to add.

AlexKusnezov:
@VossenTech my workaround ist tu use **optimistic**, that way brightness, color and temp will show in home assistant, here's my config:
```
- schema: json
  name: "Deckenleuchte (Studio)"
  unique_id: deckenleuchte_studio
  command_topic: "milight/command/0x1F0F/rgb_cct/1"
  state_topic: "milight/state/0x1F0F/rgb_cct/1"
  qos: 1
  optimistic: true
  retain: true
  color_temp: true
  rgb: true
  brightness: true
  effect: true
  effect_list:
    - night_mode
    - white_mode
    - '0'
    - '1'
    - '2'
    - '3'
    - '4'
    - '5'
    - '6'
    - '7'
    - '8'
  device:
    configuration_url: "http://192.168.2.102/"
    connections: [["ESPMH", "0x1F0F"]]
    identifiers: "0x1F0F"
    manufacturer: "Lighteu"
    model: "LT-NGU10-01 (FUT103)"
    name: "Deckenleuchte (Studio)"
```

works perfect:
![Screenshot_20230507-144159_Home Assistant](https://user-images.githubusercontent.com/33514406/236678347-e278c11a-272d-42a9-a80d-f6bd2eed5f8a.png)
![Screenshot_20230507-144213_Home Assistant](https://user-images.githubusercontent.com/33514406/236678361-1c9937d3-9e76-4a0d-8d20-e3ee2dec8f0e.png)
[...]

PR #800, opened 2023-09-08
Add support for color_mode state field for use with HomeAssistant

sidoh:
relevant issue: #792


sidoh:
this was merged into v1.11.0.

PR #853, opened 2024-11-23
Add a setting to ignore some protocols when listening

### LINKED ISSUE #845: Add ability to listen only to a specific protocol
## TL;DR
I would like a setting to select a specific remote type for listening, which improves receive reliability to nearly 100% from ~50%.

## Description

After I got the FUT092 remote working, I noticed that the listening is not very reliable. Sometimes I have to press 3-4 times before it shows up in the sniffer. It is much less reliable than the light is actually reacting. I enabled the debug printing and got these outputs and similar:

```
Radio is available
Packet received (12 bytes): 09 C4 14 70 0A 9A 04 D0 99 15 8C 2B
Successfully parsed packet of length 10
Received packet (len = 10)!
NRF24MiLightRadio - received packet!
NRF24MiLightRadio - Checking packet length (expecting 10, is 10)
Packet id: 50197
Testing whether formater for ID 32 can handle packet: with protocol ID 37...
Testing whether formater for ID 37 can handle packet: with protocol ID 37...
Radio is available
Packet received (9 bytes): 09 C4 14 70 0A 9A 04 D0 99 
Failed CRC: expected 4480, got 99D0
Radio is available
Packet received (9 bytes): 09 C4 14 70 0A 9A 04 D0 99 
Failed CRC: expected 4480, got 99D0
Radio is available
Packet received (9 bytes): 09 C4 14 70 0A 9A 04 D0 99
Failed CRC: expected 4480, got 99D0
```
These are all retries from the same button press on the remote.

As you can see, the packets are getting cut off:
Packet received (12 bytes): **09 C4 14 70 0A 9A 04 D0 99** 15 8C 2B
Packet received (9 bytes): **09 C4 14 70 0A 9A 04 D0 99** 

In this case, the retries by the remote were enough to catch the correct one, but other times it missed the packet entirely. First I thought there is some receiver interference that drops part of the packets, but the packet length is not determined by the data at all. I dug in the code to see what is causing this and found that the radio receiver is reconfigured constantly between the different protocol types and packet lengths. So depending on the exact timing, it might be decoded correctly, or it might be missed.

I think this is a good default state, because it allows to detect any kind of remote after pressing repeatedly. However, I only own one remote and would like to use it for automation. So for me, reliability is more important than supporting a range of devices. This is why I propose a UI setting to select a certain remote type from the list, or listen for all of them. After I removed the switching, the remote was detected every time.

## UI Mockup
This is a quick mockup for the settings:
image

Jojo-1000:
I added the `ignored_listen_protocols` setting to filter out unnecessary protocols. This improves the receive reliability as discussed in the linked issue.

Closes #845 

I decided to use an ignore list rather than a whitelist because any new protocols will always be enabled by default.

Screenshot:
image

Maybe it is not really fitting with the repeats category, it could also be put under the listen channel select.

Also please check my openapi changes, I have no experience with how to use that.


web2/api/api-zod.ts
```diff
@@ -576,30 +582,6 @@ const TransitionData = TransitionArgs.and(
     .passthrough()
 );
 const postTransitions_Body = TransitionData.and(BulbId);
-const PacketMessage = z
```
sidoh, line 579: Make sure to run the generator with `--export-schemas`. This gets used.

sidoh:
Looks good! openapi spec changes look right. In general if it produces the right types for the frontend it's fine.

There are some integration tests written in ruby that should probably be updated with the new setting. I'm happy to take that, though. 

Might be worth creating a new section for "listening." Can put listen repeats, listen channel and this new setting under there.

Jojo-1000:
I made a new section "📡 Listening" and fixed the generated parts

sidoh:
Great! Thanks very much for the PR!

